PR #5, opened 2015-04-07
Added in webjars and tried to read the application.properties using Dyna...

mbucknell:
...micReadOnly Properties. I have left it out for now.

@isuftin , in index.jsp on line 20 if I try to use propsFile as a parameter to the DynamicReadOnlyProperties instantiation, the app hangs and never loads and never shows an error. I'm not sure what I'm doing wrong.


mbucknell:
@ivan, @jiwalker-usgs  set me straight. I'll fix this up more tomorrow. Still want to put the reading of dynamic variables in a base.jsp and probably the common links and script that would be used on any page.

pom.xml
```diff
@@ -63,6 +63,15 @@
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		
+		<!-- js lib versions -->
+        <jquery.version>1.11.2</jquery.version>
```
isuftin: Isn't this a pretty old version of JQuery? They're on 2.1.3 at this point.

pom.xml
```diff
@@ -63,6 +63,15 @@
...
+        <bootstrap.version>3.3.4</bootstrap.version>
```
isuftin: My one request here is to use general to specific left to right naming schemes. So.. "version.bootstrap", "version.backbone", etc. 

It makes things easier to find when eyeballing it in a more complex project.

pom.xml
```diff
@@ -119,9 +128,85 @@
...
+
+			<dependency>
+				<groupId>javax.servlet</groupId>
+				<artifactId>jstl</artifactId>
+				<version>1.2</version>
+			</dependency>
+			
+			<!-- Log4j dependencies -->
+        <!-- ================================================================== -->
+        <dependency>
```
isuftin: I'd suggest:
```
		<dependency>
			<groupId>ch.qos.logback</groupId>
			<artifactId>logback-classic</artifactId>
			<scope>compile</scope>
		</dependency>
		<dependency>
			<groupId>org.slf4j</groupId>
			<artifactId>log4j-over-slf4j</artifactId>
			<scope>compile</scope>
		</dependency>
		<dependency>
			<artifactId>slf4j-api</artifactId>
			<groupId>org.slf4j</groupId>
		</dependency>
		<dependency>
			<groupId>org.slf4j</groupId>
			<artifactId>slf4j-simple</artifactId>
		</dependency>
```

We've moved away from going directly against log4j

src/main/webapp/index.jsp
```diff
@@ -1,9 +1,48 @@
...
+				URL applicationProperties = getClass().getClassLoader().getResource("application.properties");
+				File propsFile = new File(applicationProperties.toURI());
+				props = new DynamicReadOnlyProperties(propsFile);
+				props = props.addJNDIContexts(new String[0]);
+			} catch (Exception e) {
+			}
+			}
+
+		String applicationVersion = props.get("application.version");
+		String versionJquery = props.get("version.jquery");
```
isuftin: I notice you're using my suggested general -> specific naming scheme here :smile:

src/main/webapp/index.jsp
```diff
@@ -1,9 +1,48 @@
+
 <%@page contentType="text/html" pageEncoding="UTF-8"%>
+
+
 <!DOCTYPE html>
 <html>
+
     <head>
+		
+		<%@page import="java.io.File"%>
```
isuftin: These imports should be at the very top of this file

pom.xml
mbucknell: Going to ask Blodgett if it's okay not to support IE 8.

src/main/webapp/index.jsp
mbucknell: Actually planning on putting those in a separate jsp so it can be included in multiple pages.

src/main/webapp/index.jsp
```diff
@@ -1,9 +1,48 @@
...
+		%>
+		
         <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
-        <title>JSP Page</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <title>GDP Page</title>
+		<link type="text/css" rel="stylesheet" href="webjars/bootstrap/<%=versionBootstrap%>/css/bootstrap.min.css" />
+		<link type="text/css" rel="stylesheet" href="webjars/bootstrap/<%=versionBootstrap%>/css/bootstrap-responsive.min.css" />
+		<link type="text/css" rel="stylesheet" href="webjars/font-awesome/<%=versionFontAwesome%>/css/font-awesome.min.css" />
+		<script type="text/javascript" src="webjars/jquery/<%= versionJquery%>/jquery.min.js"></script>
```
isuftin: Check out https://github.com/isuftin/coastal-hazards/blob/master/coastal-hazards-portal/src/main/webapp/index.jsp#L124

I'd suggest we let a JNDI property tell us whether we're on development or production and choose a minified or non-minified version of our and third-party javascript and css based on that. 

Also, we will want to pull in a compressor maven plugin ( https://github.com/isuftin/coastal-hazards/blob/master/coastal-hazards-portal/pom.xml#L242-L260 ) so we can also minify our stuff for production.

src/main/webapp/index.jsp
isuftin: Good point. I should have done that in other projects I'm on

src/main/webapp/META-INF/context.xml
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
isuftin: Please remove this file

mbucknell:
Initial wiring of the application. @cschroed-usgs  and @isuftin, I'm going to go ahead and merge.

PR #11, opened 2015-04-08
Adding the data details satellite page. Or 'How to add a satellite page'

cschroed-usgs:


cschroed-usgs:
@isuftin and @mbucknell , for your reference.

src/main/webapp/js/advanced/templates/datadetail.html
```diff
@@ -0,0 +1 @@
+<h1>DEEETAILS</h1>
```
cschroed-usgs: Note the elaborate content.

PR #12, opened 2015-04-08
Templating

mbucknell:
Modified the util templateLoader to pass in the directory where to find templates. Changed BaseView to require a template option be specified. This decouples the dependancy of BaseView on the templateLoader that previously existed.


mbucknell:
@cschroed-usgs , go ahead and merge if it's acceptable.

cschroed-usgs:
:+1:

PR #17, opened 2015-04-09
Datadetails initial markup

cschroed-usgs:


src/main/webapp/js/advanced/templates/datadetail.html
```diff
@@ -1 +1,30 @@
...
+	    <option value="dummy2">dummy2</option>
+	    <option value="dummy3">dummy3</option>
+	    <option value="dummy4">dummy4</option>
+	</select>
+    </div>
+    <div class="form-horizontal">
+    <div class="form-group">
+	<label for="start-date" class="col-xs-12 col-sm-4 text-center">Date Range</label>
+	<input id="start-date" class="col-xs-12 col-sm-3" type="date" required/>
+	<label for="end-data" class="col-xs-12 col-sm-1" >to</label>
```
isuftin: "end-data" should be "end-date"

PR #18, opened 2015-04-09
Initial pass of loading config object

isuftin:


src/main/webapp/js/advanced/init.js
```diff
@@ -12,11 +15,20 @@ $(document).ready(function() {
 	];
 
 	var PARTIALS = [];
-
+	
 	GDP.ADVANCED.templates = GDP.util.templateLoader('js/advanced/templates/');
 	var loadTemplates = GDP.ADVANCED.templates.loadTemplates(TEMPLATES);
 	var loadPartials = GDP.ADVANCED.templates.registerPartials(PARTIALS);
-	$.when(loadTemplates, loadPartials).always(function() {
+	var loadConfigModel = $.when.apply(null, [$.ajax('config', {
```
isuftin: Should this be hidden away or do you think this is fine here?

src/main/webapp/js/advanced/init.js
```diff
@@ -12,11 +15,20 @@ $(document).ready(function() {
...
+			success: function (data) {
+				GDP.config = new GDP.model.Config(data);
+			},
+			error: function () {
+				// TODO: What to do here?
```
isuftin: Thoughts?

src/main/webapp/WEB-INF/web.xml
```diff
@@ -4,13 +4,22 @@
...
 	</servlet-mapping>
+	
+	<servlet>
+		<servlet-name>config</servlet-name>
+		<jsp-file>/WEB-INF/jsp/config.jsp</jsp-file>
+	</servlet>
+	<servlet-mapping>
+		<servlet-name>config</servlet-name>
+		<url-pattern>/config</url-pattern>
+	</servlet-mapping>
```
mbucknell: I thought we were dropping the config page.

src/main/webapp/js/advanced/init.js
cschroed-usgs: Could you please clarify your question?

src/main/webapp/js/advanced/init.js
mbucknell: Do we have reasonable defaults for everything? If not this should show an alert or error dialog.

src/main/webapp/js/advanced/init.js
cschroed-usgs: One suggestion: to keep it simple you could call $.when($.ajax(..)) here instead of $.when.apply

src/main/webapp/WEB-INF/web.xml
isuftin: We are. This is a routing for an ajax endpoint to get back json. This is never used by a human

src/main/webapp/js/advanced/init.js
isuftin: Just looks like ugly code bunched in with clean code. OCD

src/main/webapp/js/advanced/init.js
mbucknell: You can set defaults in the model.

src/main/webapp/js/advanced/init.js
isuftin: Ah. Ok. Pretty rusty on $.promises. I will try that

src/main/webapp/js/advanced/init.js
cschroed-usgs: Maybe an error modal. http://getbootstrap.com/javascript/#modals

isuftin: Yeah, right now the only thing I'm pulling from the server is "isDevelopment". I will default to defaults on err

mbucknell:
Okay I think I understand what you are doing now. :+1:

cschroed-usgs:
:+1:

PR #20, opened 2015-04-09
Creating Process models and a Processes collection

isuftin:


src/main/webapp/js/advanced/collections/Processes.js
```diff
@@ -0,0 +1,9 @@
+/*jslint browser: true*/
+/*global Backbone*/
+var GDP = GDP || {};
+
+GDP.collection = GDP.collection || {};
```
mbucknell: I think want to namespace it as GDP.Advanced.collection.

src/main/webapp/js/advanced/models/Process.js
```diff
@@ -0,0 +1,7 @@
+/*jslint browser: true*/
+/*global Backbone*/
+var GDP = GDP || {};
+
+GDP.model = GDP.model || {};
+
+GDP.model.Process = Backbone.Model.extend({});
```
mbucknell: Same comment about namespacing here.

src/main/webapp/js/advanced/collections/Processes.js
isuftin: ok. is it GDP.ADVANCED, GDP.Advanced or GDP.advanced?

isuftin: Looks like ADVANCED

PR #21, opened 2015-04-09
Gdp 857

mbucknell:


mbucknell:
@isuftin, One of us is going to have some merging to do. I did not update config.jsp for endpoint.geoserver

PR #23, opened 2015-04-09
Process model

cschroed-usgs:
There's some overlap with @isuftin's stuff here. Not sure how to proceed. Thoughts?

### DIFF
--- a/src/main/webapp/js/advanced/models/DataSourceModel.js
+++ b/src/main/webapp/js/advanced/models/DataSourceModel.js
@@ -1,0 +1,27 @@
+/*jslint browser: true*/
+/*global Backbone*/
+var GDP = GDP || {};
+
+GDP.model = GDP.model || {};
+
+GDP.model.DataSourceVariable = Backbone.Model.extend({
+    defaults:{
+	name: null
+    }
+});
+
+GDP.model.DataSourceVariables = Backbone.Collection.extend({
+    model: GDP.model.DataSourceVariable,
+    initialize: function(models, options){
+	this.bind("add", options.view.dataSourceVariableAdd);
+    }
+});
+
+GDP.model.DataSource = Backbone.Model.extend({
+    defaults:{
+	startDate: null,
+	endDate: null,
+	url: null,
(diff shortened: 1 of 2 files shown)


src/main/webapp/js/advanced/models/DataSourceModel.js
```diff
@@ -0,0 +1,27 @@
...
+GDP.model.DataSourceVariable = Backbone.Model.extend({
+    defaults:{
+	name: null
+    }
+});
+
+GDP.model.DataSourceVariables = Backbone.Collection.extend({
+    model: GDP.model.DataSourceVariable,
+    initialize: function(models, options){
+	this.bind("add", options.view.dataSourceVariableAdd);
```
mbucknell: Is options.view supposed to be a Backbone.View object? I don't think it's a good practice to have a model dependent on a view. The view can use "listenTo" to list for changes in a model.

cschroed-usgs: Aha! jackpot! Thank you!

mbucknell:
OCD thing: I don't like naming files with a name that is already in the directory name, i.e. ProcessModels which is in models.

I'm not sure what the overlap is with Ivan's code. I think you should talk to him

mbucknell:
Take back previous comment about file naming. I've looked at this some more. I'm still a little confused about what is a ProcessModel. But at some point we need a model which contains all of the models that are directly associated with our views.

isuftin:
In my mind, a ProcessModel is a description of a WPS Processing Algorithm. Carl, is that what you see it as?

cschroed-usgs:
Ivan, I was writing it as, well to steal from Mary, "a model which contains all of the models that are directly associated with our views". Basically, it's the hub's model. I thought of the WPS Processes as 'Algorithms', but I think it's better to call them 'Processes' like you've done. Maybe we should call the hub's model a 'Job' instead -- I think it makes sense to say that as we move through the wizard we are constructing a Job to submit to the GDP.

cschroed-usgs:
I'm going to close this as it's getting mixed up with my page-specific stuff.

PR #25, opened 2015-04-09
Start of event handling for the area of interest, attribute, and values ...

mbucknell:
...selection. Code works the first time, but don't try to change any of the drop down values. That doesn't work.

### DIFF
--- a/src/main/webapp/js/advanced/views/SpatialView.js
+++ b/src/main/webapp/js/advanced/views/SpatialView.js
@@ -8,8 +8,41 @@ GDP.view = GDP.view || {};
 
 GDP.view.SpatialView = GDP.util.BaseView.extend({
 
+	events : {
+		'change #select-aoi' : 'changeName',
+		'change #select-attribute' : 'changeAttribute',
+		'change #select-values' : 'changeValues'
+	},
+
+	initialize : function(options) {
+
+		GDP.util.BaseView.prototype.initialize.apply(this, arguments);
+
+		this.getAvailableFeatures().then(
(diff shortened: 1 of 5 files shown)


mbucknell:
I'm going to merge so that my changes get in and I don't have to deal with merge nightmares. Go ahead and comment but know that I'm still working on the update* functions .

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -8,29 +8,126 @@ GDP.view = GDP.view || {};
...
+	changeName : function(ev) {
+		this.model.set('name', ev.target.value);
+	},
+
+	changeAttribute : function(ev) {
+		this.model.set('attribute', ev.target.value);
+	},
+
+	changeValues : function(ev) {
+		var values = [];
```
cschroed-usgs: This method can be one line: 
```
this.model.set('values', _.pluck(ev.target.selectedOptions, 'value');
```

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -8,29 +8,126 @@ GDP.view = GDP.view || {};
...
+			{
+				request : 'GetCapabilities'
+			},
+			false,
+			populateFeatureTypesSelectBox
+		);
+	},
+
+	changeName : function(ev) {
+		this.model.set('name', ev.target.value);
```
cschroed-usgs: Thank you for providing good examples of how to do event handling.

PR #26, opened 2015-04-10
Algorithm selection description

isuftin:


src/main/webapp/js/advanced/templates/process.html
```diff
@@ -3,8 +3,16 @@
 		<button class="btn btn-default dropdown-toggle" type="button" id="button-dropdown-algorithm" data-toggle="dropdown" aria-expanded="true">Alg Button<span class="caret"></span></button>
 		<ul class="dropdown-menu" role="menu" aria-labelledby="button-dropdown-algorithm">
 			{{#each this}}
-			<li role="presentation"><a role="menuitem" tabindex="-1" href="#">{{this.attributes.title}}</a></li>
```
isuftin: Using an href that links to a standalone hash messes with backbone routing and brings me back to the root of the context I'm in (which in my case is /advanced#/). This functionality works without using the hash href, so that's ok. I would still like to know a decent workaround for this with backbone.

isuftin:
Carl, please take a look

src/main/webapp/js/advanced/templates/process.html
cschroed-usgs: What I'm doing [here](https://github.com/USGS-CIDA/gdp_web/blob/master/src/main/webapp/js/advanced/templates/datadetail.html#L4) is
```
<form class="panel-body" action="#/">
```
Which works for some reason, though by the router rules, it shouldn't.
I suggest we change the router so that the hub is at "#/"

isuftin: What's the hub at currently? I thought it was #/ ?

isuftin:
Merging now. Got another PR on the way.

src/main/webapp/js/advanced/templates/process.html
cschroed-usgs: It's possible I don't understand the routing yet, but it looks like it is at '#' https://github.com/USGS-CIDA/gdp_web/blob/master/src/main/webapp/js/advanced/controller/AdvancedRouter.js#L14

isuftin: So if we change that to '/', will that work I wonder?

cschroed-usgs: I misunderstood the routing -- looks like a '/' is explicitly prepended to each entry, so the hub's current route of '' (empty string) can be accessed via '#/' hrefs/form actions.

PR #30, opened 2015-04-10
Now loading DataSource and Job models

isuftin:


src/main/webapp/js/advanced/init.js
```diff
@@ -44,6 +44,7 @@ $(document).ready(function() {
 			return new GDP.ADVANCED.model.Process(m);
 		}));
 		GDP.ADVANCED.router = new GDP.ADVANCED.controller.AdvancedRouter();
+		GDP.ADVANCED.Job = new GDP.ADVANCED.model.Job();
```
isuftin: We can probably find a cleaner way of doing this. I did this for expediency since I need the Job model to finish my job :D

cschroed-usgs:
:+1:

src/main/webapp/js/advanced/init.js
cschroed-usgs: That works. Could also convert it into a Singleton. My next PR will have an example of that.

PR #31, opened 2015-04-10
Switched DataSourceModel to a singleton. Implementing some inter-field d...

cschroed-usgs:
...ependencies.

### DIFF
--- a/src/main/webapp/advanced.jsp
+++ b/src/main/webapp/advanced.jsp
@@ -47,6 +47,7 @@
 		<script type="text/javascript" src="js/ogc/wfs.js"></script>
 		<script type="text/javascript" src="js/advanced/models/Process.js"></script>
 		<script type="text/javascript" src="js/advanced/models/SpatialModel.js"></script>
+		<script type="text/javascript" src="js/advanced/models/DataSourceModel.js"></script>
 		<script type="text/javascript" src="js/advanced/collections/Processes.js"></script>
 		<script type="text/javascript" src="js/advanced/views/HubView.js"></script>
 		<script type="text/javascript" src="js/advanced/views/SpatialView.js"></script>
--- a/src/main/webapp/js/advanced/controller/AdvancedRouter.js
+++ b/src/main/webapp/js/advanced/controller/AdvancedRouter.js
@@ -32,7 +32,8 @@ GDP.ADVANCED.controller.AdvancedRouter = Backbone.Router.extend({
 
 	datadetail : function() {
 		this.showView(GDP.view.DataDetailsView, {
-			template : GDP.ADVANCED.templates.getTemplate('datadetail')
+			template : GDP.ADVANCED.templates.getTemplate('datadetail'),
+			model: GDP.ADVANCED.model.dataSource
 		});
 	},
 
--- a/src/main/webapp/js/advanced/models/DataSourceModel.js
+++ b/src/main/webapp/js/advanced/models/DataSourceModel.js
@@ -1,26 +1,32 @@
 /*jslint browser: true*/
 /*global Backbone*/
 var GDP = GDP || {};
+(function(){
+    "use strict";
+    GDP.ADVANCED = GDP.ADVANCED || {};
 
-GDP.ADVANCED = GDP.ADVANCED || {};
+    GDP.ADVANCED.model = GDP.ADVANCED.model || {};
 
-GDP.ADVANCED.model = GDP.ADVANCED.model || {};
+    GDP.ADVANCED.model.DataSourceVariable = Backbone.Model.extend({
+	defaults:{
+	    name: null
+	}
+    });
 
-GDP.ADVANCED.model.DataSourceVariable = Backbone.Model.extend({
-    defaults:{
-	name: null
-    }
-});
+    GDP.ADVANCED.model.DataSourceVariables = Backbone.Collection.extend({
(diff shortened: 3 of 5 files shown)


src/main/webapp/js/advanced/controller/AdvancedRouter.js
```diff
@@ -32,7 +32,8 @@ GDP.ADVANCED.controller.AdvancedRouter = Backbone.Router.extend({
 
 	datadetail : function() {
 		this.showView(GDP.view.DataDetailsView, {
-			template : GDP.ADVANCED.templates.getTemplate('datadetail')
+			template : GDP.ADVANCED.templates.getTemplate('datadetail'),
+			model: GDP.ADVANCED.model.dataSource
```
cschroed-usgs: reference to a singleton

src/main/webapp/js/advanced/models/DataSourceModel.js
```diff
@@ -1,26 +1,32 @@
...
+    //singleton
 
-GDP.ADVANCED.model.DataSource = Backbone.Model.extend({
-    defaults:{
-	startDate: null,
-	endDate: null,
-	url: null,
-	variables: new GDP.ADVANCED.model.DataSourceVariables
-    }
-});
```
cschroed-usgs: Here I create the class, but hide it using function scope. An instance of the class is exposed globally via the namespace. Note it is lower-cased because it isn't a type, it's an instance.

src/main/webapp/advanced.jsp
```diff
@@ -47,6 +47,7 @@
 		<script type="text/javascript" src="js/ogc/wfs.js"></script>
 		<script type="text/javascript" src="js/advanced/models/Process.js"></script>
 		<script type="text/javascript" src="js/advanced/models/SpatialModel.js"></script>
+		<script type="text/javascript" src="js/advanced/models/DataSourceModel.js"></script>
```
isuftin: You may want to merge my stuff first and then see if you have collisions since I already did add this in my PR

src/main/webapp/js/advanced/views/DataDetailsView.js
```diff
@@ -1,8 +1,59 @@
+/*jslint browser: true*/
+/*global _ jQuery*/
 var GDP = GDP || {};
+(function(_, $){
+    "use strict";
```
cschroed-usgs: This immediately invoked anonymous function is a good way to enforce strict mode for everything in a file.

src/main/webapp/js/advanced/models/DataSourceModel.js
```diff
@@ -1,26 +1,32 @@
...
\ No newline at end of file
+    var DataSource = Backbone.Model.extend({
+	defaults:{
+	    startDate: null,
+	    endDate: null,
+	    url: null,
+	    variables: new GDP.ADVANCED.model.DataSourceVariables
+	}
+    });
+    GDP.ADVANCED.model.dataSource = new DataSource();
```
mbucknell: I'm not a big fan of doing this with a singleton. I'm wondering what the implications are on unit testing. Another way of creating the models that an app needs is to create the models in the router.

cschroed-usgs: I agree. In the future we should move to a more testable approach. I am using singleton patterns right now because it minimizes the number of moving parts while I figure out Backbone's moving parts. I am having trouble getting the hub view to render content from the child models and an absolute global reference to a singleton object is the least complicated way to debug it.

src/main/webapp/js/advanced/views/DataDetailsView.js
cschroed-usgs: @isuftin, sorry I didn't elaborate on this. This is a "good way" to enforce strict mode for everything in a file because it is concatenation-safe. Typing "use strict" outside of a function scope makes it hard to concatenate your scripts for production -- one "use strict" statement outside of function scope makes the browser interpret everything in the concatenated file in strict mode. [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Strict_mode_for_scripts) has more info.

isuftin: @cschroed-usgs I get what you're doing here. However, doing it this way does also force you to architect your application a bit differently due to immediate execution. I don't have an opinion on that for this specific object, but it may be crucial in other objects that need to be initialized after others, right? Or am I interpreting the impact too severely?

cschroed-usgs: There shouldn't be any impact on the way you architect your application. The contents of your file will be evaluated by the JS engine when loaded whether or not your file's contents are wrapped in an immediately invoked anonymous function. There is no difference.

isuftin: Right, but we may need to wait for document.ready for some things, no? Just shooting darts here.

isuftin: Actually, thinking about this, I think this is fine. I agree that this is a good way of doing this.

cschroed-usgs: Yep, you can still put a call to document.ready inside of the IIAF for event-driven stuff.

cschroed-usgs: * You got it :)

PR #35, opened 2015-04-10
Rendering some of the nested model content in the hub...

cschroed-usgs:
... page.


src/main/webapp/js/advanced/views/HubView.js
```diff
@@ -3,6 +3,13 @@ var GDP = GDP || {};
 GDP.view = GDP.view || {};
 
 GDP.view.HubView = GDP.util.BaseView.extend({
+    model: GDP.ADVANCED.model.job,
+    render: function(){
+	this.$el.html(this.template());
+	var url = this.model.get('dataSource').get('url');
```
cschroed-usgs: url is always null, even after you go into the data details page and specify the url.

When debugging,
this.model.get('dataSource') !== GDP.ADVANCED.model.dataSource
...but GDP.ADVANCED.model.dataSource is a singleton, so how can that be? Cloning? If so, why?

src/main/webapp/js/advanced/views/DataDetailsView.js
```diff
@@ -17,12 +17,24 @@ var GDP = GDP || {};
...
+	}
+    };
+    var urlPicker = {
+	selector: '#data-source-url',
+	$el: null
+    };
+    var urlChanged = function(ev){
+	var value = ev.target.value;
+	//todo: validation
+	this.model.set('url', value);
```
cschroed-usgs: this.model === GDP.ADVANCED.model.dataSource

src/main/webapp/js/advanced/views/HubView.js
```diff
@@ -3,6 +3,14 @@ var GDP = GDP || {};
 GDP.view = GDP.view || {};
 
 GDP.view.HubView = GDP.util.BaseView.extend({
```
cschroed-usgs: @mbucknell , I got child model content to render in the hub view by going hardcore singleton like you suggested. I am not super happy about this either, but I'm glad we have a proof of concept.

mbucknell:
After talking to Carl and doing a little googling, I think we should either consider using a plugin like Backbone.nestedtypes http://volicon.github.io/backbone.nestedTypes/ or flattening the model so that there is only a single model JobModel containing all the attributes needed to create a job. Each view would use the attributes that are appropriate. Thoughts?

isuftin:
I thought we were only using a single JobModel that every satellite and hub accesses. No?

mbucknell:
Yes ... the question is whether that model consists of other models (the view specific ones we have created) or whether contains the attributes needed for all satellites.

isuftin:
Ah. I was under the impression that each view holds its own model(s) but also updates the Job singleton when needed and can also watch the Job model for changes to update its own model.

cschroed-usgs:
> isuftin... and can also watch the Job model for changes to update its own model. ...
The hub view summarizes state. I don't think it is supposed to permit modification of state. Accordingly, if we go with the current approach of a Job Model with nested models for each satellite, there is no need for satellite views/models to watch the Job Model.

cschroed-usgs:
Mary, thanks for investigating Model design solutions. 

 > Backbone.nestedtypes
It looks ok. I'm not crazy about the psuedo-static typing stuff, but I think it's worth a look-see.

 > flattening the model
It might be more Backbone kosher, but it's not a super modular approach. As a consequence it makes me worry about merge conflicts -- everyone has legitimate motivation to modify the same file. I favor eschewing flattening if we can help it.

I'm going to merge this. Going forward I'd like to experiment with both approaches that Mary proposed.

PR #39, opened 2015-04-13
Removed the SpatialModel and put attributes into the common JobModels.js...

mbucknell:
.... Fixed SpatialView to  properly handle updates to the attribute values on the view.


cschroed-usgs:
:+1:

PR #41, opened 2015-04-13
Enforcing IDE indent policy in the pom.

cschroed-usgs:


cschroed-usgs:
I think we're all using NetBeans on this project. If it's cool with both of you I'd like to enforce an indent policy so that we can stop seeing tab<->spaces diffs.

mbucknell:
Fine with me.

isuftin:
Please leave IDE specific configuration out of the source code. Especially the pom. Either share a configuration file amongst ourselves, add it to the project and gitignore it or set it at the IDE level for each person working on the project.

cschroed-usgs:
Ivan, you assert what we should do without providing justification for why we should do it. Why? In general I don't want most IDE-specific configuration checked into source because it references resources that are individual to developers' workstations. I believe this case is exceptional because we are all using the same IDE, and we gain clarity in code review by enforcing workstation-invariant IDE configuration. The solutions you propose permit each developer to (forget to) have any indentation policy they want, and forces the rest of the team to waste time looking at the consequent messy diffs.

 > Sharing a configuration file amongst ourselves, add it to the project and gitignore it 

That will fall out of sync. Devs will lose the file. New devs won't have the file. Space<->tabs diffs will ensue. This could be somewhat remedied by following the example of exampleContext.xml and context.xml. To be precise, we would check an example-nb-configuration.xml into source and tell new devs to copy it to nb-configuration.xml on clone, and tell existing devs to copy it whenever it is updated. This still suffers from the problem of individual devs being able to override the policy by modifying the indentation settings in their IDE's GUIs. Changes are written to nb-configuration.xml. The developer doesn't have to face the music of his/her noncompliance because the file is .gitignored.

 > set it at the IDE level for each person working on the project

This is essentially the same as the above, but with more opportunity for human error in clicking the right buttons in the GUI instead of reading them in from a file. This approach will fall out of sync even faster. Devs delete and re-clone repos, losing the nb-config file in the process. Since there is no file checked into the repo enforcing or describing an indentation policy, devs have no convenient means to get on the same page as the other devs on their team. Devs will erroneously apply settings from one project to another. Space<->tabs diffs will ensue.

isuftin:
I'd like to keep this open source project agnostic of things like IDE configuration. I see things like that  as a pollution to the code base. 

We also can't guarantee that any future developers will be using NetBeans which renders this configuration moot.

Perhaps we can make use of a tooling-agnostic solution like https://maven.apache.org/plugins/maven-checkstyle-plugin/check-mojo.html

Thoughts?

mbucknell:
I'm going to close this pull request without merging.

PR #42, opened 2015-04-14
Gdp 857

mbucknell:
Major changes. Added the map. Added model change event handlers for the aoiName and aoiValues to draw the selected features of interest on the map. I am using a WMS request mostly because that's what the old GDP app did.  The majority of the interesting changes are in js/util/mapUtils.js and js/advanced/views/SpatialView.

Still to do would be to add tests for these new features. Next step will be to add the upload shapefile functionality.

### DIFF
--- a/src/main/webapp/js/advanced/views/SpatialView.js
+++ b/src/main/webapp/js/advanced/views/SpatialView.js
@@ -3,6 +3,7 @@
 /*global $*/
 /*global GDP.util.BaseView*/
 /*global GDP.util.SelectMenuView*/
+/*global GDP.util.mapUtils*/
 /*global GDP.OGC.WFS*/
 
 var GDP = GDP || {};
@@ -20,8 +21,28 @@ GDP.ADVANCED.view.SpatialView = GDP.util.BaseView.extend({
 		'change #select-values' : 'changeValues'
 	},
 
+	render : function() {
+		GDP.util.BaseView.prototype.render.apply(this, arguments);
+		this.map.render('spatial-map');
+		this.map.zoomToExtent(this.map.getMaxExtent(), true);
+	},
+
 	initialize : function(options) {
 		"use strict";
+		var baseLayers = [GDP.util.mapUtils.createWorldStreetMapLayer()];
+		var controls = [
+			new OpenLayers.Control.Navigation(),
+			new OpenLayers.Control.MousePosition({
+				prefix: 'POS: ',
+				numDigits: 2,
+				displayProjection: GDP.util.mapUtils.WGS84_GEOGRAPHIC
+			}),
+			new OpenLayers.Control.ScaleLine({
+				geodesic: true
+			}),
+			new OpenLayers.Control.Zoom()
+		];
+		this.map = GDP.util.mapUtils.createMap(baseLayers, controls);
 
 		GDP.util.BaseView.prototype.initialize.apply(this, arguments);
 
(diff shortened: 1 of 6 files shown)


src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -12,146 +13,242 @@ GDP.ADVANCED = GDP.ADVANCED || {};
...
-		var name = this.model.get('aoiName');
-
-		this.model.set('aoiAttributeValues', []);
-		this.attributeValuesSelectMenuView.$el.val(null);
-		this.attributeValuesSelectMenuView.updateMenuOptions([]);
-
-		if ((name) && (attribute)) {
-			GDP.OGC.WFS.callWFS(
+			this.updateAOILayer();
+			this.listenTo(this.model, 'change:aoiName', this.updateAttributes);
```
isuftin: I wonder if backbone allows this to be made into a JSON object? 

this.listenTo(this.model, { k1 : f1, k2 : f2, k3 : f3 ... }) ? Just wondering.

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -12,146 +13,242 @@ GDP.ADVANCED = GDP.ADVANCED || {};
...
+			this.attributeValuesSelectMenuView.$el.val(null);
+			this.attributeValuesSelectMenuView.updateMenuOptions([]);
+
+			if ((name) && (attribute)) {
+				GDP.OGC.WFS.callWFS(
+					{
+						request : 'GetFeature',
+						typename : name,
+						propertyname : attribute,
+						maxFeatures : 5001 //TODO verify that this is correct
```
isuftin: Perhaps set as top level k:v in this object?

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -12,146 +13,242 @@ GDP.ADVANCED = GDP.ADVANCED || {};
...
+					},
+					false,
+					_.bind(function(data) {
+						var optionValues = [];
+
+						$(data).find(attribute).each(function() {
```
isuftin: Simplify with _.filter($(data).find(attribute), function (){ ... }) ?

cschroed-usgs: Good idea

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -12,146 +13,242 @@ GDP.ADVANCED = GDP.ADVANCED || {};
...
+					{
+						request : 'DescribeFeatureType',
+						typename : name
+					},
+					false,
+					_.bind(function(data) {
+						var optionValues = [];
+
+						$(data).find('complexContent').find('element[name!="the_geom"]').each(function() {
+							optionValues.push($(this).attr('name'));
```
cschroed-usgs: Not a blocker, but this could be simplified with a _.map

cschroed-usgs:
:+1:

src/main/webapp/js/advanced/views/SpatialView.js
mbucknell: I don't think so. The event object is a string. You can specify multiple events in the same string ... we could do something like this.listenTo(this.model, 'change:param1 change:param2', this.updateAction).

src/main/webapp/js/advanced/views/SpatialView.js
mbucknell: I think is specific to this particular function. If we were always setting maxFeatures to a number it might make more sense.

src/main/webapp/js/advanced/views/SpatialView.js
mbucknell: Except I'm filtering and mapping. I think I'll leave this one as is.

PR #44, opened 2015-04-15
Gdp 857

mbucknell:
I added the ability to zoom to the selected feature. Things to note:

I added a new function to wfs to get the bounds from the cache. Also reformated the file.

### DIFF
--- a/src/main/webapp/js/ogc/wfs.js
+++ b/src/main/webapp/js/ogc/wfs.js
@@ -6,43 +6,43 @@ var GDP = GDP || {};
 GDP.OGC = GDP || {};
 
 
-GDP.OGC.WFS = (function() {
-    var capabilitiesCache;
+GDP.OGC.WFS = (function () {
+	var capabilitiesCache;
 
-    function _callWFS(data, async, successCallback) {
-        var defaultData = {
-            'service': 'WFS',
-            'version': '1.1.0'
-        }
+	function _callWFS(data, async, successCallback) {
+		var defaultData = {
+			'service': 'WFS',
+			'version': '1.1.0'
+		}
 
-        var wfsData = {};
+		var wfsData = {};
 
-        // Merge defaultData with data, putting results in wfsData. If there are
-        // any conflicts, the property from data will overwrite the one in defaultData.
-        $.extend(wfsData, defaultData, data);
-        GDP.logger.debug('GDP: Calling WFS Service with a '+ wfsData.request + ' request.');
-        var promise = $.ajax({
-            url: GDP.config.get('application').endpoints.geoserver  + '/wfs',
-            async: async,
-            data: wfsData,
-            cache: false,
-            success: function(data, textStatus, jqXHR) {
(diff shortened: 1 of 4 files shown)


src/main/webapp/js/util/mapUtils.js
```diff
@@ -42,11 +42,10 @@ GDP.util.mapUtils = (function() {
 	};
 
 	that.createMap = function(layers, controls) {
-		var maxExtent = new OpenLayers.Bounds(GDP.config.get('map').extent.conus['3857']);
```
cschroed-usgs: Curiosity :cat: says "why is it better to read the extent from the config vs hardcoding it?" Not a blocker.

src/main/webapp/js/ogc/wfs.js
```diff
@@ -6,44 +6,64 @@ var GDP = GDP || {};
...
-        GDP.logger.debug('GDP: Calling WFS Service with a '+ wfsData.request + ' request.');
-        var promise = $.ajax({
-            url: GDP.config.get('application').endpoints.geoserver  + '/wfs',
-            async: async,
-            data: wfsData,
-            cache: false,
-            success: function(data, textStatus, jqXHR) {
-                if (!$(data).find('ExceptionReport').length) {
-                    if ('GetCapabilities' === wfsData.request) {
-                        GDP.OGC.WFS.cachedGetCapabilities = data;
```
cschroed-usgs: Nice elimination of global references via closures :+1:

src/main/webapp/js/util/mapUtils.js
mbucknell: It may need to be used in other places is the standard answer. Right now we are only using it here and it is in a utility package but it's possible to use it in other places too.

src/test/js/specs/SpatialView_spec.js
```diff
@@ -31,14 +32,17 @@ describe('GDP.ADVANCED.VIEW.SpatialView', function() {
 
 		wfsDeferred = $.Deferred();
 
+		spyOn(OpenLayers.Layer, 'WMS');
```
cschroed-usgs: You are a Jasmine pro. I didn't know about spyOn. Thanks for enlightening.

cschroed-usgs:
:+1:

src/main/webapp/js/util/mapUtils.js
cschroed-usgs: Makes sense. Thanks for explaining.

src/test/js/specs/SpatialView_spec.js
mbucknell: spyOn is great. You add the ability to call through, you can return a value of your choosing, or execute a function of your choosing. It is very handy and absolutely necessary for testing code that calls OpenLayers

cschroed-usgs: Awesome! Good to know.

PR #45, opened 2015-04-17
Gdp 857

mbucknell:
Added the ability to upload files. I decided not to use jasny-bootstrap in favor of jquery-fileupload. As part of this I created a new utility view, AlertView to show alert messages. Also added initialization for the view. 

Still to do are tests for AlertView and updated tests for SpatialView but the current tests do run.

### DIFF
--- a/src/main/webapp/WEB-INF/web.xml
+++ b/src/main/webapp/WEB-INF/web.xml
@@ -28,7 +28,7 @@
         <servlet-class>gov.usgs.cida.proxy.AlternateProxyServlet</servlet-class>
         <init-param>
             <param-name>forward-url-param</param-name>
-            <param-value>gdp.endpoint.geoserver</param-value>
+            <param-value>gdp.geoserver.endpoint</param-value>
         </init-param>
         <init-param>
             <param-name>readTimeout</param-name>
--- a/src/main/webapp/WEB-INF/web.xml
+++ b/src/main/webapp/WEB-INF/web.xml
@@ -39,4 +39,107 @@
         <servlet-name>GeoServerProxyServlet</servlet-name>
         <url-pattern>/geoserver/*</url-pattern>
     </servlet-mapping>
+	
+	<servlet>
+        <servlet-name>shapefile-upload-servlet</servlet-name>
+        <servlet-class>gov.usgs.cida.geoutils.geoserver.servlet.ShapefileUploadServlet</servlet-class>
+        <!--
+		Used for reading JNDI variables. All JNDI variables used for this application
+		will be prefixed with this parameter. Example:
+		<Environment name="coastal-hazards.max.upload.file.size" type="java.lang.String" value="20000" />
+		-->
+        <init-param>
+            <param-name>application.name</param-name>
+            <param-value>gdp</param-value>
+        </init-param>
+        <!--
+		Used to limit the size of incoming files measured in bytes. [Optional]
+		0 = Integer.MAX_VALUE
+		-->
+        <init-param>
+            <param-name>max.upload.file.size</param-name>
+            <param-value>0</param-value>
+        </init-param>
+        <!--
+		The URL of the geoserver that will be used for uploading.
+		-->
+        <init-param>
+            <param-name>geoserver.endpoint</param-name>
+            <param-value>http://localhost:8081/geoserver/</param-value>
+        </init-param>
+        <!--
+		The username used to authenticate against when writing against GeoServer.
+		-->
+        <init-param>
+            <param-name>geoserver.username</param-name>
+            <param-value>admin</param-value>
+        </init-param>
+        <!--
+		Workspace name that files will be uploaded to by default. [Optional. If this
+		parameter is not provided here, in JNDI and during the request, upload will fail]
+		-->
+        <init-param>
+            <param-name>default.upload.workspace</param-name>
+            <param-value>upload</param-value>
+        </init-param>
+        <!--
+		Store name that files will be uploaded to by default. [Optional. If this
+		parameter is not provided here, in JNDI and during the request, upload will fail]
+		-->
+        <init-param>
+            <param-name>default.upload.storename</param-name>
+            <param-value/>
+        </init-param>
+        <!--
+		Default SRS is not defined. If a SRS name is not passed to during the
+		request, the request will fail [Optional. If this parameter is not provided
+		here, in JNDI and during the request, upload will fail]
+		-->
+        <init-param>
+            <param-name>default.srs</param-name>
+            <param-value>EPSG:4326</param-value>
+        </init-param>
+        <!--
+		The key to search for in the upload form post to find the file
+		[Optional. If this parameter is not provided "qqfile" will be used]
+		-->
+        <init-param>
+            <param-name>filename.param</param-name>
+            <param-value/>
+        </init-param>
+        <!--
+		Should the existing layer, if one exists, be overwritten?
+		[Optional. If this parameter is not provided "false" will be used]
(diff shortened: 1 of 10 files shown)


src/main/webapp/js/util/AlertView.js
```diff
@@ -0,0 +1,40 @@
...
+		render : function() {
+			this.$el.append(this.template({
+				message : this.message,
+				alertClass : this.alertClass
+			}));
+		},
+
+		initialize : function(options) {
+			this.$el = $(options.el);
+			var html = '<div class="alert {{alertClass}} alert-dismissable" role="alert">' +
```
isuftin: No external resource for this? :)

mbucknell: Not that I'm aware of. These are bootstrap alerts. Admittedly I did not look hard.

src/main/webapp/advanced.jsp
```diff
@@ -43,10 +44,13 @@
 		
 		<%@include file="/WEB-INF/jsp/scripts.jsp" %>
 		<script type="text/javascript" src="webjars/openlayers/<%= versionOpenLayers%>/OpenLayers<%= development ? "" : ".debug"%>.js"></script>
-
+		<script type="text/javascript" src="webjars/jquery-ui/<%= versionJqueryUI%>/jquery-ui<%= development ? "" : ".min"%>.js"></script>
+		<script type="text/javascript" src="webjars/jquery-file-upload/<%= versionJqueryFileUpload%>/js/jquery.fileupload.js"></script>
+		
 		<script type="text/javascript" src="js/util/templateLoader.js"></script>
```
isuftin: I'm going to give a suggestion (not related to this PR). You should:

1- Minify your JS and select whether or not you're using your minified JS based on a 'development' JNDI property) like we're doing on like 47 here. 

2- Along with that, use application version as part of the minified filename as in:

a) https://github.com/USGS-CIDA/coastal-hazards/blob/master/coastal-hazards-portal/src/main/webapp/index.jsp#L30

b) https://github.com/USGS-CIDA/coastal-hazards/blob/master/coastal-hazards-portal/src/main/webapp/index.jsp#L139-L162

src/main/webapp/js/util/AlertView.js
isuftin: Oh I just meant: Did you not want to create an external html file for this?

mbucknell: I don't think so. There isn't anything that's really app specific here. I did a similar thing for SelectMenuView.

isuftin: Ok, I can get down with that line for partitioning what is an external file and what is in-line

src/main/webapp/advanced.jsp
mbucknell: Thanks and will do ... but not until after Carl is done with his ticket and I'm done with mine.

src/main/webapp/WEB-INF/web.xml
```diff
@@ -39,4 +39,107 @@
...
+        <init-param>
+            <param-name>max.upload.file.size</param-name>
+            <param-value>0</param-value>
+        </init-param>
+        <!--
+		The URL of the geoserver that will be used for uploading.
+		-->
+        <init-param>
+            <param-name>geoserver.endpoint</param-name>
+            <param-value>http://localhost:8081/geoserver/</param-value>
```
cschroed-usgs: I has confuse: is this url supposed to be hardcoded, or read in from config?

src/main/webapp/advanced.jsp
cschroed-usgs: Interesting. Thanks for transferring the CCH knowledge Ivan!

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -20,13 +20,30 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
 		events : {
 			'change #select-aoi' : 'changeName',
 			'change #select-attribute' : 'changeAttribute',
-			'change #select-values' : 'changeValues'
+			'change #select-values' : 'changeValues',
```
cschroed-usgs: Might want to kill that trailing comma.

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -51,20 +68,65 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
...
+
+						if (warning) {
+							this.alertView.show('alert-warning', 'Upload succeeded with warning' + warning);
+						}
+						else {
+							this.alertView.show('alert-success', 'Upload was successful.');
+						}
+
+						this.getAvailableFeatures().then(
+							_.bind(function() {
```
cschroed-usgs: I don't use _.bind much, but I don't know why. The way I usually accomplish this is to assign 'this' to a variable in the outer function, and access the variable from the inner function via closure.
```
function(){
var self = this;
this.getAvailableFeatures().then(function(){
self.model.set()
...
});
}
```

I'm not sure what the tradeoffs are for each approach. What do you think?

src/main/webapp/WEB-INF/web.xml
isuftin: I think this is coming in as a default, though I might be incorrect

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -75,11 +137,25 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
...
+				this.model.set('aoiAttribute', attribute);// Need to
+				this.updateSelectedAoiAttribute();
+			}
+			// Need to reset the aoiAttributeValues because updateSelectedAOIAttribute clears it.
+			if (values.length !== 0) {
+				this.model.set('aoiAttributeValues', values);
+				this.updateSelectedAoiAttributeValues();
+			}
+
+			this.listenTo(this.model, 'change:aoiName', this.updateSelectedAoiName);
```
cschroed-usgs: Ahhh interesting. I didn't anticipate that we would have to listen to the model, but it totally makes sense here since the uploader callback is modifying the model without triggering a jquery event that we can hook into via the View's 'events' property.

src/main/webapp/js/advanced/views/SpatialView.js
mbucknell: I'll clean these up on the next go around.

cschroed-usgs:
:+1:

src/main/webapp/js/advanced/views/SpatialView.js
mbucknell: I like using bind rather then assigning this to a variable ... This may just be personal preference. I don't see either ways as being better or worse.

src/main/webapp/js/advanced/views/SpatialView.js
cschroed-usgs: :+1:

src/main/webapp/js/advanced/views/SpatialView.js
mbucknell: This also separates an action which updates the model from a model change updating the DOM.

src/main/webapp/WEB-INF/web.xml
mbucknell: That is the value. We are proxying the geoserver (see line 40 in the pom.xml).

cschroed-usgs: Gotcha. Makes sense.

PR #46, opened 2015-04-17
Data details with web service calls

cschroed-usgs:
This is not done, but it's been a while since my last merge.


src/main/webapp/WEB-INF/jsp/scripts.jsp
```diff
@@ -8,5 +8,8 @@
 <script type="text/javascript" src="webjars/underscorejs/<%= versionUnderscore%>/underscore<%= development ? "" : "-min"%>.js"></script>
 <script type="text/javascript" src="webjars/handlebars/<%= versionHandlebars%>/handlebars<%= development ? "" : ".min"%>.js"></script>
 <script type="text/javascript" src="webjars/backbonejs/<%= versionBackbone%>/backbone<%= development ? "" : ".min"%>.js"></script>
+<script type="text/javascript" src="webjars/openlayers/<%= versionOpenLayers%>/OpenLayers<%= development ? "" : ".debug"%>.js"></script>
```
mbucknell: I'd like to keep page specific things like OpenLayers, FileUpload, and bootstrap-datepicker out of scripts.jsp since it will be included in all pages. Eventually this application will grow and other pages are not likely to need all of these.

src/main/webapp/js/advanced/controller/AdvancedRouter.js
```diff
@@ -35,9 +35,11 @@ GDP.ADVANCED.controller.AdvancedRouter = Backbone.Router.extend({
 	},
 
 	datadetail : function() {
-		this.showView(GDP.view.DataDetailsView, {
+		this.showView(GDP.ADVANCED.view.DataDetailsView, {
 			template : GDP.ADVANCED.templates.getTemplate('datadetail'),
-			model: this.jobModel
+			model: this.jobModel,
+			wps: GDP.OGC.WPS(GDP.logger),
+			wpsEndpoint: GDP.config.get('application').endpoints.utilityWps
```
mbucknell: I like passing these things into the view ... it should make testing easier. I might change SpatialView to follow the same pattern.

src/main/webapp/js/advanced/views/DataDetailsView.js
```diff
@@ -23,66 +23,313 @@ var GDP = GDP || {};
...
-    GDP.view.DataDetailsView = GDP.util.BaseView.extend({
-	events : (function(){
+    GDP.ADVANCED.view.DataDetailsView = GDP.util.BaseView.extend({
+	'events' : (function(){
 	    var ret = {};
-	    ret['change ' + variablePicker.selector] = variableSelected;
-	    ret['change ' + urlPicker.selector] = urlChanged;
+	    ret['change ' + variablePicker.selector] = 'selectVariables';
+	    ret['change ' + urlPicker.selector] = 'changeUrl';
+		ret['changeDate ' + datePickers.start.selector] = 'changeStartDate';
```
mbucknell: Fix indentation.

src/main/webapp/js/advanced/views/DataDetailsView.js
```diff
@@ -23,66 +23,313 @@ var GDP = GDP || {};
...
-	this.model.get('dataSourceVariables').reset(variables);
-	this.render();
-    };
+	var VARIABLE_WPS_PROCESS_ID = 'gov.usgs.cida.gdp.wps.algorithm.discovery.ListOpendapGrids';
+	var DATE_RANGE_WPS_PROCESS_ID = 'gov.usgs.cida.gdp.wps.algorithm.discovery.GetGridTimeRange';
     
-    GDP.view.DataDetailsView = GDP.util.BaseView.extend({
-	events : (function(){
+    GDP.ADVANCED.view.DataDetailsView = GDP.util.BaseView.extend({
+	'events' : (function(){
```
mbucknell: Is there a reason you used a function invocation rather than just assigning an object to events?

src/main/webapp/js/advanced/views/DataDetailsView.js
```diff
@@ -23,66 +23,313 @@ var GDP = GDP || {};
...
+		var variables = _.map(ev.target.options, function (option) {
+			return {
+				'text': option.text,
+				'value': option.value,
+				'selected': option.selected
+			};
+		});
+
+		this.model.get('dataSourceVariables').reset(variables);
+		this.render();
```
mbucknell: Does re rendering the entire view make sense? I'm wondering if the rerendering is visible to the user. It seems heavy handed for updating a single variable to rerender the whole page.

isuftin: Is it possible to re-render scoped only to the variable?

mbucknell: What I have been doing is updating the model when an action updates an attribute. I then have a listener on the model's attributes that handles updating the DOM for that particular attribute.

isuftin: Could you link to where in the source you have that? That may be helpful to Carl to do that here.

mbucknell: SpatialView has examples of this, https://github.com/USGS-CIDA/gdp_web/blob/master/src/main/webapp/js/advanced/views/SpatialView.js

src/main/webapp/js/advanced/views/DataDetailsView.js
```diff
@@ -23,66 +23,313 @@ var GDP = GDP || {};
...
+			var minDate = this.model.get('minDate');
+			endDatePicker.datepicker('setStartDate', minDate);
+		}
 	}
-    });
+});
 
-}(_, jQuery));
+GDP.ADVANCED.view.DataDetailsView.VARIABLE_WPS_PROCESS_ID = VARIABLE_WPS_PROCESS_ID;
+GDP.ADVANCED.view.DataDetailsView.DATE_RANGE_WPS_PROCESS_ID = DATE_RANGE_WPS_PROCESS_ID;
```
mbucknell: Are these used outside of this view? If not, should they be name spaced global variables.

src/main/webapp/js/ogc/wps.js
```diff
@@ -191,3 +232,4 @@ var WPS = function () {
         }
     };
 };
+}());
```
mbucknell: Nice refactor of wps.js. I would like to do the same for wfs.js at some point.

src/main/webapp/WEB-INF/jsp/scripts.jsp
cschroed-usgs: Good call. I will relocate them.

mbucknell:
Nice tests! I haven't used runs and waitsFor. Typically I just manipulate the deferreds in the code, but I'll have to consider using this in the future. I know you are still working on this so feel free to go ahead and merge (after moving the <script> statements).

:+1:

src/main/webapp/js/advanced/controller/AdvancedRouter.js
cschroed-usgs: Thanks. I adopted this approach for exactly that reason -- so I could mock dependencies during testing. I also switched from the revealing module pattern to the module pattern so that I could reliably patch public-facing members during testing if needed. Relevant commit: https://github.com/cschroed-usgs/gdp_web/commit/1512110d1f892f9a9a85295d7ee39b0f59192569

src/main/webapp/js/advanced/views/DataDetailsView.js
cschroed-usgs: That's weird. Will do.

src/main/webapp/js/advanced/views/DataDetailsView.js
cschroed-usgs: Yes. Throughout this class I am implementing a single source of truth for my selectors so that it will be easy for the code to react should the selectors need to change. The Backbone event object wants selectors in the property names. I need to use an expression to access the single source for each selector. In object literal notation, you cannot use an expression in a property name, you can only use a string. Since I need to use an expression to access the single source for the selector, I need to use square bracket notation. This immediately-invoked function expression ([IIFE](http://benalman.com/news/2010/11/immediately-invoked-function-expression/)) lets me use square bracket notation. The IIFE returns a backbone-kosher events object while maintaining a single source of truth for selectors. I think this is a pretty trivial layer of indirection to add for the future-proofing that it buys you. If both of you disagree, I can back out and replace the variables with hardcoded strings, and the function with an object literal.

mbucknell: Ok that's a good reason. I do think it makes reading the code more difficult.  It's been my observation that selector names don't change, so I'm not as opposed to hardcoding the strings in the few places that I am using them. @isuftin , do you have an opinion?

src/main/webapp/js/advanced/views/DataDetailsView.js
cschroed-usgs: Thanks for focusing on this guys. I was going to bring it up if y'all didn't :) The approach I've implemented is a pretty heavy-handed page-wide render. The re-rendering is not visible to the user, but it is certainly wasting some client resources parsing and interpolating the template every time. On the flip side I think it reduces coupling by sacrificing control over which specific thing is updated in the UI. I'm conflicted on this. I am not sure I like putting selectors in model listeners either because it increases coupling between the view and the template, even if there is a single source of truth for all selectors within the class.

I'm open to suggestions and additional discussion, but if we don't make headway today I suggest we merge this and revisit it when we have more Backbone-fu to inform our design decisions.

mbucknell: I don't have a problem putting selectors in the model listeners. In fact for the Spatial page, I wouldn't want to re-render because it would mean making re-rendering the map along with the associated wms and wfs calls.

I think you can go ahead and merge but I do think this is worth talking about.

src/main/webapp/js/advanced/views/DataDetailsView.js
cschroed-usgs: Good call. They are not used outside of this view. I will remove these exports to the global space.

src/main/webapp/js/ogc/wps.js
cschroed-usgs: Thanks! Switching to a promise-based interface was a little more work than I thought initially, but it simplified testing of the view class a lot and enabled us to enforce consistent WPS-level error message parsing. The WPS class still lacks tests, but like the view, now that it uses promises it should be easier to test too

cschroed-usgs:
Thanks Mary. The last message was cut off. Could you please repeat what I should make sure to do prior to merging?

mbucknell:
after moving the script tags from scripts.jsp to advanced.jsp.

cschroed-usgs:
Roger that and Wilco :)

src/main/webapp/js/advanced/views/DataDetailsView.js
cschroed-usgs: Thanks Mary. I don't know if the OpenLayers widget case points us clearly toward a technique. We can both accomplish something like that in the approaches that we are using. For the approach I've been taking:

```
GDP.ADVANCED.view.SomeView = GDP.util.BaseView.extend({
    //...,
    'map' : false,
    
    'render':function(){
        //conditionally render map
        if(false === this.map){
            var mapDiv = this.$el.find('#some-map');
            this.map = new OpenLayers.Map(mapDiv)'
        }
        //always render the other stuff
        this.$el.find('#some-non-map').html(this.template({
            'property' : this.model.get('blah'),
            //...
        }));
    }
    //...
```

mbucknell: This is putting more logic into render which I don't like. And in my case, it isn't just the map.

cschroed-usgs:
Fixed the flukes, so I'm going to merge. Let's keep thinking about the events IIFE and the rendering approaches.

src/main/webapp/js/advanced/views/DataDetailsView.js
cschroed-usgs: I discovered that the way I'm doing it actually does have some negative user impacts -- when the user loads a bunch of variables, scrolls down the list and ctrl+selects another one, they jump to the top of the list. This is a potentially very annoying UX for someone picking sparse variables that wants to go through the list top-to-bottom. I'm going to switch to an approach more like Mary's.

PR #47, opened 2015-04-20
Added tests for AlertView and added a few more tests for SpatialView.

mbucknell:
I'm going to declare this ticket done. To do's would be refactoring wfs.js and improving the test coverage.

### DIFF
--- a/src/test/js/specs/AlertView_spec.js
+++ b/src/test/js/specs/AlertView_spec.js
@@ -1,0 +1,46 @@
+describe('GDP.util.AlertView', function() {
+
+	beforeEach(function() {
+		$('body').append('<div id="test-div"></div>');
+	});
+
+	afterEach(function() {
+		$('#test-div').remove();
+	});
+
+	it('Expects that initialize creates the view but does not render it', function() {
+		var testView = new GDP.util.AlertView({
+			el : '#test-div'
+		});
+
+		expect(testView.$el).toEqual($('#test-div'));
+		expect(testView.template).toBeDefined();
+	});
+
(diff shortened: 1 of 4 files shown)


src/test/js/specs/AlertView_spec.js
```diff
@@ -0,0 +1,46 @@
...
+
+	beforeEach(function() {
+		$('body').append('<div id="test-div"></div>');
+	});
+
+	afterEach(function() {
+		$('#test-div').remove();
+	});
+
+	it('Expects that initialize creates the view but does not render it', function() {
```
cschroed-usgs: How does this test verify that the view hasn't rendered?

src/test/js/specs/AlertView_spec.js
```diff
@@ -0,0 +1,46 @@
...
+
+		testView.show('alert-danger', 'Danger message');
+		expect(testView.template).toHaveBeenCalledWith({
+			message : 'Danger message',
+			alertClass : 'alert-danger'
+		});
+		expect($('#test-div .alert').length).toBe(1);
+	});
+
+	it('Expects that a second invocation of show removes the first alert and replaces with the seconde', function() {
```
cschroed-usgs: seconde -> second

src/test/js/specs/AlertView_spec.js
mbucknell: It doesn't, but I can fix that! Thanks for the catch.

cschroed-usgs:
Looks good! Nice tests. I'm glad we've got a tested alert solution. After those small fixes you are :+1: to merge.

PR #50, opened 2015-04-22
Fixed bug that @mbucknell discovered. Responses that only describe one v...

cschroed-usgs:
...ariable are now marked as valid and correctly parsed.


cschroed-usgs:
@mbucknell, thanks for reporting this bug!

PR #51, opened 2015-04-23
Gdp 859

mbucknell:
I decided to put all of the processVariable in there own model which is why JobModel has had many attributes eliminated and replaced by processVariables. I think this makes sense since process variables change passed on algorithm selected and it makes it convenient to handle as a group.

That was probably the biggest change besides hooking things together to save/recover attributes on the process page.

### DIFF
--- a/src/main/webapp/js/advanced/controller/AdvancedRouter.js
+++ b/src/main/webapp/js/advanced/controller/AdvancedRouter.js
@@ -44,8 +44,9 @@ GDP.ADVANCED.controller.AdvancedRouter = Backbone.Router.extend({
 	},
 
 	process : function() {
-		this.showView(GDP.view.ProcessView, {
+		this.showView(GDP.ADVANCED.view.ProcessView, {
 			template : GDP.ADVANCED.templates.getTemplate('process'),
+			algorithmTemplate : GDP.ADVANCED.templates.getTemplate('algorithm-config'),
 			model: this.jobModel
 		});
 	},
--- a/src/main/webapp/js/advanced/init.js
+++ b/src/main/webapp/js/advanced/init.js
@@ -21,13 +21,16 @@ $(document).ready(function() {
 	var PARTIALS = [];
 
 	GDP.ADVANCED.templates = GDP.util.templateLoader('js/advanced/templates/');
-	
+
 	var loadConfigModel = $.when($.ajax('config', {
 			success: function (data) {
 				GDP.config = new GDP.model.Config(data);
+			},
+			error : function (jqXHR, textStatus) {
+				console.log('Can not read config ' + textStatus);
 			}
 		}));
-	
+
 	// I need to load up my config model since one of the views I load depends on it
 	// Load up the process collection based on incoming model definitions from the config object
 
@@ -38,7 +41,7 @@ $(document).ready(function() {
 		initializeLogging({
 			LOG4JS_LOG_THRESHOLD: GDP.config.get('application').development === 'true' ? 'debug' : 'info'
 		});
-		
+
 		GDP.ADVANCED.templates.registerHelpers();
 		GDP.logger = log4javascript.getLogger();
 		var jobModel = new GDP.ADVANCED.model.Job();
--- a/src/main/webapp/js/advanced/views/AlgorithmConfigView.js
+++ b/src/main/webapp/js/advanced/views/AlgorithmConfigView.js
@@ -6,118 +6,121 @@ var GDP = GDP || {};
 GDP.ADVANCED = GDP.ADVANCED || {};
 GDP.ADVANCED.view = GDP.ADVANCED.view || {};
 
-GDP.ADVANCED.view.AlgorithmConfigView = GDP.util.BaseView.extend({
+(function() {
+	"use strict";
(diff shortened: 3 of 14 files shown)


src/main/webapp/js/advanced/controller/AdvancedRouter.js
```diff
@@ -44,8 +44,9 @@ GDP.ADVANCED.controller.AdvancedRouter = Backbone.Router.extend({
 	},
 
 	process : function() {
-		this.showView(GDP.view.ProcessView, {
+		this.showView(GDP.ADVANCED.view.ProcessView, {
```
isuftin: I feel like we should not have gone with this naming scheme (no "ADVANCED"). It doesn't sit well with me and I think it might be because it's tied too close to the fractured nature of the application (advanced vs not advanced). Maybe it's just me. Anyway, no need to change anything at this point, just getting that out there.

dblodgett-usgs: Mary and I discussed this... the plan is to finish the initial work on the app then remove ADVANCED for something more suitable. Great minds...

src/main/webapp/js/advanced/init.js
```diff
@@ -21,13 +21,16 @@ $(document).ready(function() {
...
 
 	GDP.ADVANCED.templates = GDP.util.templateLoader('js/advanced/templates/');
-	
+
 	var loadConfigModel = $.when($.ajax('config', {
 			success: function (data) {
 				GDP.config = new GDP.model.Config(data);
+			},
+			error : function (jqXHR, textStatus) {
+				console.log('Can not read config ' + textStatus);
```
isuftin: Because this is a severe error (and only of of many possible severe errors that can happen on init), we should find a graceful way to let the user know they're not going anywhere.

Will our app have a splash screen? If so, we should show an error with a contact email there and not advance beyond. If no splash screen, what else can we do to show the user that they're hosed?

src/main/webapp/js/advanced/views/AlgorithmConfigView.js
```diff
@@ -1,22 +1,141 @@
...
 		return this;
+	},
+
+	/*
+	 * Hook for updating text process variable
+	 * @param {Jquery event } ev
+	 */
+	changeTextProcessVariable : function(ev) {
+		"use strict";
+		var key = ev.target.id.replace('input-', '');
```
isuftin: This code is repeated multiple times in this file. I wonder if we can refactor this in the future?

src/main/webapp/js/advanced/views/ProcessView.js
```diff
@@ -1,71 +1,107 @@
...
+
+			this.listenTo(this.model, 'change:algorithmId', this.displayAlgorithmDescription);
+		},
+
+		render : function () {
+			var algorithmId = this.model.get('algorithmId');
+
+			// Hiding the rendered template until the selected algorithm description is shown so that
+			// there isn't flashing when an algorithm has already been selected
+			this.$el.hide();
```
isuftin: Doesn't this happen almost instantaneously? Does it not cause flicker?

src/main/webapp/js/util/templateLoader.js
```diff
@@ -66,8 +66,8 @@ GDP.util.templateLoader = function(templateDir) {
 				}
 				return options.inverse(this);
 			},
-			'defaultChecked' : function (bool) {
-				if (bool) {
+			'defaultChecked' : function (boolStr) {
```
isuftin: Will this always be a string? I tried to code for truthiness. 

'true' === true -> false

isuftin:
:+1: Looks good to go

src/main/webapp/js/advanced/init.js
mbucknell: I agree. I put in the console.log after having typos in config.jsp and going down the wrong path. I think once we get to a point were the app is more fully functioning and we understand how people come into it, we can address this.

src/main/webapp/js/advanced/views/ProcessView.js
mbucknell: It's a problem when you come into the app with an algorithmId already defined. The template was rendered and you could see all of the algorithm's descriptions and then all put one goes away.

src/main/webapp/js/advanced/views/AlgorithmConfigView.js
mbucknell: I'll refactor it now as this is easy to do.

src/main/webapp/js/advanced/views/ProcessView.js
isuftin: Ah! Didn't think of that. Good point

isuftin:
Hijacking this pull request to ask if you think introducing LESS at this early stage is a consideration for you? I don't want to push it as I'm not actively on here right now but am seeding the idea for you.

mbucknell:
I think so. Obviously we haven't really dealt with page styling yet. One of the tickets for this iteration that I'll work on next week is to make a "real" hub page showing what parameters have been set. I think that would be a good time to start it up. Thanks for giving me the push!

PR #58, opened 2015-05-13
Gdp 934

mbucknell:
This adds the draw polygon feature which required some refactoring of the job process creation code so that draw: shapefiles don't send a filter with the web processing request. The draw feature itself was pretty much lifted from the old code and updated to use deferred rather than callbacks.

There was some other minor refactoring for the xml parsing leftover from GDP-935.

### DIFF
--- a/src/main/webapp/js/advanced/models/JobModel.js
+++ b/src/main/webapp/js/advanced/models/JobModel.js
@@ -3,7 +3,7 @@
 /*global _*/
 var GDP = GDP || {};
 (function(){
-	"use strict;";
+	"use strict";
 
     GDP.ADVANCED = GDP.ADVANCED || {};
 
@@ -189,13 +189,10 @@ var GDP = GDP || {};
 					cql_filter : GDP.util.mapUtils.createCQLFilter(attribute, values),
 					maxFeatures : 5001
 				}, 'POST').done(function(data) {
-					// parse gml ids from result. I have to have two selectors, one with the namespace and one without.
-					var name_tag = name.substr(name.indexOf(':') + 1);
-					var esc_name = name.replace(':', '\\:');
 					var result = [];
-					($(data).find(esc_name + ', ' + name_tag).each(function() {
+					GDP.util.findXMLNamespaceTags($(data), name).each(function() {
 						result.push($(this).attr('gml:id'));
-					}));
+					});
 					deferred.resolve(result);
 				}).fail(function() {
 					GDP.logger.error('Get Selected features failed');
--- a/src/main/webapp/js/advanced/models/JobModel.js
+++ b/src/main/webapp/js/advanced/models/JobModel.js
@@ -43,6 +43,14 @@ var GDP = GDP || {};
 		},
 
 		/*
+		 * @returns {Boolean}
+		 */
+		needsAoiAttributeValues : function() {
+			var namespace = this.get('aoiName').split(':')[0];
+			return namespace !== 'draw';
+		},
+
+		/*
 		 * Returns the process model for the model's algorithmId. If nothing matches, will return undefined
 		 * @returns {GDP.ADVANCED.model.Process}
 		 */
@@ -138,12 +146,14 @@ var GDP = GDP || {};
 				return (new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()))).toISOString();
 			};
 
+			var aoiAttribute = this.get('aoiAttribute');
+
 			var deferred = $.Deferred();
 
 			var getDataSourceUrl = this.getWCSDataSourceUrl();
 
 			var result = {
-				FEATURE_ATTRIBUTE_NAME : [this.get('aoiAttribute')],
+				FEATURE_ATTRIBUTE_NAME : [aoiAttribute ],
 				DATASET_ID : _.map(this.getSelectedDataSourceVariables(), function(model) {
 					return model.get('value');
 				}),
@@ -179,7 +189,10 @@ var GDP = GDP || {};
 			var values = this.get('aoiAttributeValues');
 
 			var deferred = $.Deferred();
-			if ((name) && (attribute) && (values.length > 0)) {
+			if (!this.needsAoiAttributeValues()) {
+				deferred.resolve([]);
+			}
+			else if ((name) && (attribute) && (values.length > 0)) {
 				GDP.OGC.WFS.callWFS({
 					request : 'GetFeature',
 					typename : name,
@@ -244,11 +257,11 @@ var GDP = GDP || {};
                   'xsi:schemaLocation="http://www.opengis.net/wfs ../wfs/1.1.0/WFS.xsd"> ' +
 				  '<wfs:Query typeName="' + name + '" ' + ((srs) ? 'srsName="' + srs + '"' : '') + '> ' +
 				  '<wfs:PropertyName>' + (geomProperty ? geomProperty : 'the_geom') + '</wfs:PropertyName> ' +
-                 '<wfs:PropertyName>' + attribute + '</wfs:PropertyName>';
+                 '<wfs:PropertyName>' + (attribute ? attribute :  'ID') + '</wfs:PropertyName>';
 
 			var deferred = $.Deferred();
 			this.getSelectedFeatureIds().done(function(ids) {
-				if (ids[0] !== '*') {
+				if (ids.length > 0) {
 					result += '<ogc:Filter>';
 					_.each(ids, function(id) {
 						result += '<ogc:GmlObjectId gml:id="' + id + '"/>';
@@ -272,7 +285,7 @@ var GDP = GDP || {};
 
 
 			}
-			else if (!(this.get('aoiAttribute')) || !(this.get('aoiAttributeValues'))) {
+			else if (this.needsAoiAttributeValues() && (!(this.get('aoiAttribute')) || !(this.get('aoiAttributeValues')))) {
 				result.push('Select a feature within the area of interest.');
 			}
 			return result;
--- a/src/main/webapp/js/advanced/models/JobModel.js
+++ b/src/main/webapp/js/advanced/models/JobModel.js
@@ -1,6 +1,7 @@
 /*jslint browser: true*/
 /*global Backbone*/
 /*global _*/
+/*global Handlebars*/
 var GDP = GDP || {};
 (function(){
 	"use strict";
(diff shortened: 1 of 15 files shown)


src/main/webapp/js/advanced/models/JobModel.js
```diff
@@ -244,11 +243,11 @@ var GDP = GDP || {};
                   'xsi:schemaLocation="http://www.opengis.net/wfs ../wfs/1.1.0/WFS.xsd"> ' +
 				  '<wfs:Query typeName="' + name + '" ' + ((srs) ? 'srsName="' + srs + '"' : '') + '> ' +
 				  '<wfs:PropertyName>' + (geomProperty ? geomProperty : 'the_geom') + '</wfs:PropertyName> ' +
-                 '<wfs:PropertyName>' + attribute + '</wfs:PropertyName>';
+                 '<wfs:PropertyName>' + (attribute ? attribute :  'ID') + '</wfs:PropertyName>';
```
isuftin: We should probably be building this via handlebars at this point

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -209,23 +195,127 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
...
+		 * @returns {undefined}
+		 */
+		toggleDrawControl : function() {
+			var $toggle = $('#draw-polygon-btn');
+			var turnDrawOn = !$toggle.hasClass('active');
+
+			this.drawFeatureLayer.setVisibility(turnDrawOn);
+			if (turnDrawOn) {
+				this.drawFeatureControl.activate();
+				$('#draw-polygon-div').show();
```
isuftin: Bring this up as a variable near $toggle

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -209,23 +195,127 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
...
+			}
+		},
+
+		/*
+		 * Saves the drawn polygon after verifying that all necessary information has been entered.
+		 * @returns {undefined}
+		 */
+		saveDrawnPolygons : function() {
+			var self = this;
+			var name = $('#polygon-name-input').val();
```
isuftin: Make variable from $('#polygon-name-input'), used multiple times

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -209,23 +195,127 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
...
+			this.drawFeatureLayer.protocol.setFeatureType(name);
 
+			// Create the a datastore for the new shapefile
+			var wpsInputs = {
+				name : [name]
+			};
+			var wpsOutputs = ['layer-name'];
+			this.wps.sendWpsExecuteRequest(
+				GDP.config.get('application').endpoints.utilityWps + '/WebProcessingService',
+				'gov.usgs.cida.gdp.wps.algorithm.filemanagement.CreateNewShapefileDataStore',
```
isuftin: We should probably have these in a top-level config object

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -335,27 +432,199 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
...
-					this.highlightLayer.mergeNewParams({
-						layers : name,
-						cql_filter : filter
-					});
-				}
-				else {
-					this.highlightLayer = GDP.util.mapUtils.createAOIFeaturesLayer(name, filter);
-					this.map.addLayer(this.highlightLayer);
+			if (name) {
+				if ((attribute) && (values.length !== 0)) {
```
isuftin: Why the encapsulation in parens here?

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -335,27 +432,199 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
...
-				if (this.highlightLayer) {
-					this.highlightLayer.mergeNewParams({
-						layers : name,
-						cql_filter : filter
-					});
-				}
-				else {
-					this.highlightLayer = GDP.util.mapUtils.createAOIFeaturesLayer(name, filter);
-					this.map.addLayer(this.highlightLayer);
+			if (name) {
```
isuftin: Weird nested if. 

I see no reason not to have:
    
    if (name && (attribute && values.length !== 0)) { ... }

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -335,27 +432,199 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
...
 			}
-			else if (this.highlightLayer) {
+			if (this.highlightLayer) {
 				this.map.removeLayer(this.highlightLayer);
 				this.highlightLayer = null;
 			}
+		},
+
+		/*
+		 * Sets up
```
isuftin: Finish documentation

src/main/webapp/js/advanced/views/SpatialView.js
thongsav-usgs: May want to switch to using classes instead of ids. Using ids limits the use of this view to a single instance per page (not that you'd need more than one of these). Also avoid the risk of something else clobbering your element. 

I've also started trying to get more diligent about only grabbing DOM elements from local scope (starting at this.el) so that you ensure the view always grabs only elements within it's scope (including the scope of the sub-views).

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -335,27 +432,199 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
...
+		 * @param {Jquery element} $fileUploaderInput
+		 * @param {Jquery element} $uploadIndicator
+		 */
+		_createFileUploader : function($fileUploaderInput, $uploadIndicator) {
+			var self = this;
+			var params = {
+				maxfilesize : 167772160,
```
isuftin: OCD kicking in, single quotes around this ;D And we can also move that maxfilesize into a global config object if we have one

src/main/webapp/js/advanced/views/SpatialView.js
isuftin: If we know there's only going to be one of these, I don't see the point in using classes

thongsav-usgs: Just to be able to pick up and move the view, or turn it into a subview.

isuftin: But if you're moving the DOM and not duplicating it, I don't see how classes would be useful here. I tend to use IDs for specificity until I know that I need classes. This keeps my DOM from having untenable CSS selectors.

thongsav-usgs: Using a DOM element with an id is equivalent to initializing and using a global variable.

isuftin: I don't know what this means. Please explain.

isuftin:
I like this. It's a clean implementation with tight, functional content. There's a few comments PT and I made but overall this looks great.

src/main/webapp/js/advanced/views/SpatialView.js
thongsav-usgs: All instances of SpatialView are going to be accessing the same instance of #element-id. Further, using ids allows direct access by anyone. The assumption that SpatialView is going to be a single instance is fine, but it's really easy to directly access this div from another view/controller now, which makes it real easy to start introducing hidden coupling.

isuftin: How does making it a class make it not easy to access this div from another view/controller?

thongsav-usgs: It doesn't, you can't ever really prevent direct access, but it takes away the direct link. Another way I think about it is relative paths vs absolute paths.

mbucknell: This is not a "reusable view". It's for a specific page and in that case, I think it's appropriate to use ids rather than classes. I will start using this.$el as the root of DOM searches.

src/main/webapp/js/advanced/models/JobModel.js
mbucknell: Is it weird that we would be using a template within a model? I like the idea though.

cschroed-usgs: I like the idea of templating this xml. We should remember that introducing the templating library as a dependency might impinge on the re-usability of this client. If we don't care about that, template-it-up. If we do care, might want to stick with dumb string concatenation.

isuftin: We already are using Handlebars in the app

mbucknell: I just added a pull request for adding the xml as a static string in JobModel.js. I think this will do for now.

cschroed-usgs: @mbucknell  sounds good. @isuftin , I know this app uses handlebars. I'm talking about re-using the WFS client in other apps that might not have handlebars. I defer to Mary on whether the client is sufficiently general to be re-used by another app.

cschroed-usgs: Wait, sorry this is in JobModel. I thought we were talking about the WFS client. My bad!

src/main/webapp/js/advanced/views/SpatialView.js
mbucknell: So put them in GDP.Config as defaults? I'm not sure they need to go in config.jsp.

src/main/webapp/js/advanced/views/SpatialView.js
mbucknell: I'll fix ... it's leftover from a previous attempt at getting this function correct.

src/main/webapp/js/advanced/views/SpatialView.js
isuftin: Yeah looking back at this, Im not sure about it either. But having this gem of a WPS process name hanging around in the middle of code is a bit odd. Maybe just me

mbucknell: This isn't the only place where a WPS process name is hardcoded. I think I'll address this as part of a refactor task that I have coming up (finally changing ADVANCED to PROCESS_CLIENT) and deal with it then.

isuftin:
:confetti_ball:

PR #59, opened 2015-05-15
Gdp 205

mbucknell:
In addition to reimplementing GDP 205, I also parsed the DescribeFeature response in SpatialView to get the feature ids rather than make the call when getSelectedFeatures in JobModel was called. In that function, return the empty array if all of the feature values have been selected.

### DIFF
--- a/src/main/webapp/WEB-INF/jsp/config.jsp
+++ b/src/main/webapp/WEB-INF/jsp/config.jsp
@@ -19,12 +19,15 @@
 %>
 <%
 	Boolean development = Boolean.parseBoolean(props.getProperty("gdp.development"));
+	String maxPolygonsToShowAttributeValues = props.get("gdp.application.maxPolygonsToShowAttributeValues");
+	maxPolygonsToShowAttributeValues = (maxPolygonsToShowAttributeValues == null) ? "1000" : maxPolygonsToShowAttributeValues;
 	String geoserverEndpoint = props.get("gdp.geoserver.endpoint");
 	/* serviceEndpoints below point at the actual endpoints versus endpoints which can be proxies */
 %>
 {
 	"application" : {
 		"development" : "<%= development %>",
+		"maxPolygonsToShowAttributeValues" : <%= maxPolygonsToShowAttributeValues%>,
 		"endpoints" : {
 			"geoserver" : "geoserver",
 			"utilityWps" : "utilityWps",
--- a/src/main/webapp/js/advanced/models/JobModel.js
+++ b/src/main/webapp/js/advanced/models/JobModel.js
@@ -29,6 +29,7 @@ var GDP = GDP || {};
 	var featureTemplate = Handlebars.compile(GET_FEATURE_XML_TEMPLATE);
 
     var Job = Backbone.Model.extend({
+		SELECT_ALL_AOI_ATTRIBUTE_VALUES : ['*'],
 		defaults: {
 			//data details
 			dataSourceUrl : null,
@@ -203,7 +204,7 @@ var GDP = GDP || {};
 			var selectedFeatures;
 			var totalFeatures = [];
 
-			if ((values.length > 0 && values[0] !== '*') && (values.length < featureIds.length)) {
+			if ((values !== this.SELECT_ALL_AOI_ATTRIBUTE_VALUES) && (values.length < featureIds.length)) {
 				selectedFeatures = _.filter(featureIds, function(e) {
 					return _.contains(values, e.value);
 				});
--- a/src/main/webapp/js/advanced/models/JobModel.js
+++ b/src/main/webapp/js/advanced/models/JobModel.js
@@ -50,6 +50,7 @@ var GDP = GDP || {};
 			aoiExtent : null, // will be the extent of the layer aoiName
 			aoiAttribute : '',
 			aoiAttributeValues : [],
+			aoiAttributeFeatureIds : [], // Array of objects with {String} value property and {Array of String} ids property associated with that value.
 
 			//ows identifier for the algorithm. Ex: gov.usgs.cida.gdp.wps.algorithm.FeatureWeightedGridStatisticsAlgorithm
 			processes: new GDP.ADVANCED.collection.Processes(),
@@ -191,37 +192,20 @@ var GDP = GDP || {};
 		},
 
 		/*
-		 * Returns a promise which resolves with an array containing the feature ids that have been selected.
-		 * @returns {jquery.Deferred.promise}.
+		 * Returns an array containing the feature ids that have been selected.
+		 * @returns {Array of strings}.
 		 */
 		getSelectedFeatureIds : function() {
-			var name = this.get('aoiName');
-			var attribute = this.get('aoiAttribute');
+			var result = [];
+			var featureIds = this.get('aoiAttributeFeatureIds');
 			var values = this.get('aoiAttributeValues');
-
-			var deferred = $.Deferred();
-			if ((name) && (attribute) && (_.first(values) !== '*')) {
-				GDP.OGC.WFS.callWFS({
-					request : 'GetFeature',
-					typename : name,
-					propertyname : attribute,
-					cql_filter : GDP.util.mapUtils.createCQLFilter(attribute, values),
-					maxFeatures : 5001
-				}, 'POST').done(function(data) {
-					var result = [];
-					GDP.util.findXMLNamespaceTags($(data), name).each(function() {
-						result.push($(this).attr('gml:id'));
-					});
-					deferred.resolve(result);
-				}).fail(function() {
-					GDP.logger.error('Get Selected features failed');
-					deferred.resolve([]);
-				});
-			}
-			else {
-				deferred.resolve([]);
-			}
-			return deferred.promise();
+			var selectedFeatures = _.filter(featureIds, function(e) {
+				return _.contains(values, e.value);
+			});
+			_.each(selectedFeatures, function(e) {
+				result = result.concat(e.ids);
+			});
+			return result;
 		},
 
 		/*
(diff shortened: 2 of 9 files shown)


src/main/webapp/WEB-INF/jsp/config.jsp
```diff
@@ -19,12 +19,15 @@
 %>
 <%
 	Boolean development = Boolean.parseBoolean(props.getProperty("gdp.development"));
+	String maxPolygonsToShowAttributeValues = props.get("gdp.application.maxPolygonsToShowAttributeValues");
```
cschroed-usgs: You may be able to make this more concise by using the two-arg method of props.getProperty. Its method signature is :
```
public String getProperty(String key, String defaultValue)
```
So you could do this in one line:
```
String maxPolygonsToShowAttributeValues = props.get("gdp.application.maxPolygonsToShowAttributeValues", "1000");
```

src/main/webapp/js/advanced/models/JobModel.js
```diff
@@ -246,23 +237,17 @@ var GDP = GDP || {};
...
-			});
-			return deferred.promise();
+			var context = {
+				name : this.get('aoiName'),
+				attribute : this.get('aoiAttribute'),
+				srs : (srs) ? srs : '',
+				geomProperty : (geomProperty) ? geomProperty : '',
+				featureIds : this.getSelectedFeatureIds()
+			};
+			return featureTemplate(context);
```
cschroed-usgs: :+1:  Templating. Awww yiss.

src/main/webapp/js/advanced/models/JobModel.js
```diff
@@ -191,37 +193,26 @@ var GDP = GDP || {};
...
-					});
-					deferred.resolve(result);
-				}).fail(function() {
-					GDP.logger.error('Get Selected features failed');
-					deferred.resolve([]);
+			if ((values !== this.SELECT_ALL_AOI_ATTRIBUTE_VALUES) && (values.length < featureIds.length)) {
+				selectedFeatures = _.filter(featureIds, function(e) {
+					return _.contains(values, e.value);
+				});
+				_.each(selectedFeatures, function(e) {
```
cschroed-usgs: What do you think about using a call to _.map here instead of a call to _.each?

src/main/webapp/js/advanced/models/JobModel.js
```diff
@@ -191,37 +193,26 @@ var GDP = GDP || {};
...
-					var result = [];
-					GDP.util.findXMLNamespaceTags($(data), name).each(function() {
-						result.push($(this).attr('gml:id'));
-					});
-					deferred.resolve(result);
-				}).fail(function() {
-					GDP.logger.error('Get Selected features failed');
-					deferred.resolve([]);
+			if ((values !== this.SELECT_ALL_AOI_ATTRIBUTE_VALUES) && (values.length < featureIds.length)) {
+				selectedFeatures = _.filter(featureIds, function(e) {
```
cschroed-usgs: Seeing these back-to-back underscore function calls made me think of a feature that I'd like to explore more -- chaining:

http://underscorejs.org/#chain

What do you think of this feature? It reminds me of some examples I've seen in Clojure and in the Java 8 Streams API. I like how concise it can be.

cschroed-usgs: I guess the downside is that chaining doesn't provide you with an opportunity to set a breakpoint between the function calls and hence could make debugging a little annoying. Maybe chaining simple filter, map, reduce calls would be ok, but more complicated ones should take a breather and assign their results to variables so that the dev can inspect them via a breakpoint?

src/main/webapp/js/advanced/templates/spatial.html
```diff
@@ -44,6 +44,9 @@
 			<label for="select-values">Select Value(s)</label>
 			<select required="required" class="form-control" multiple id="select-values">
 			</select>
+			<div id="attribute-values-exceeded-div">
+				Feature has too many polygons. All will be used for processing.
```
cschroed-usgs: Maybe it will already be apparent from the context of the app, but would "All features will be used... " be clearer?

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -177,7 +177,7 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
 				});
 			}
 			else {
-				self.model.set('aoiAttributeValues', ['*']);
+				self.model.set('aoiAttributeValues', self.model.SELECT_ALL_AOI_ATTRIBUTE_VALUES);
```
cschroed-usgs: good de-duplicating use of this constant throughout the classes.

src/main/webapp/js/advanced/templates/spatial.html
cschroed-usgs: Is the text specified in the JIRA still the desired message here?
"Too many attribute values to display. All polygons will be used for processing"

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -409,29 +411,62 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
...
-						})
-					);
-
-					var optionObjects = _.map(optionValues, function(optionValue){
-						return {
-							text: optionValue,
-							value: optionValue
-						};
+					var aoiAttributeFeatureIds = [];
+					var featureInfo = GDP.util.findXMLNamespaceTags($(data), name);
```
cschroed-usgs: :+1: nice util function!

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -409,29 +411,62 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
...
+					var optionValues;
+					var optionObjects;
+					featureInfo.each(function(){
```
cschroed-usgs: Could you clarify the intent of this anonymous function by naming it and/or supplying a doc block?

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -409,29 +411,62 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
...
+								value : value,
+								ids : [id]
+							});
+						}
 					});
-					self.attributeValuesSelectMenuView.updateMenuOptions(optionObjects);
-					self.attributeValuesSelectMenuView.$el.val(optionValues);
 
+					// Limits the number of polygons that can be sent in a processing request - GDP-205
+					if (featureInfo.length < parseInt(GDP.config.get('application').maxPolygonsToShowAttributeValues)) {
```
cschroed-usgs: Do most users of the maxPolygonsToShowAttributeValues property need the value as a string or as an int? If most need it as an int, you might consider driving the call to parseInt() up to the config definition like in this pseudo code

```
config = {
...
maxPolygonsToShowAttributeValues : parseInt("<%=someJndiProp%>"),
...
}
```

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -409,29 +411,62 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
...
+						self._setVisibility(self.attributeValuesSelectMenuView.$el, true);
+						messageDiv.hide();
+						optionObjects = _.map(aoiAttributeFeatureIds, function(e){
+							return {
+								text: e.value,
+								value: e.value
+							};
+						});
+						optionValues = _.pluck(optionObjects, 'value');
```
cschroed-usgs: :+1: _.pluck ftw!

src/main/webapp/js/advanced/views/SpatialView.js
```diff
@@ -617,7 +652,14 @@ GDP.ADVANCED.view = GDP.ADVANCED.view || {};
 		 * @param {Boolean} isVisible
 		 */
 		_setVisibility : function($el, isVisible) {
-			var $inputs = $el.find(':input');
+			var $inputs;
+			if ($el.is(':input')) {
+				$inputs = $el;
+			}
+			else {
+				$inputs = $el.find(':input');
```
cschroed-usgs: Could you update the doc block to include the child-input feature too?

src/test/js/specs/JobModel_spec.js
```diff
@@ -333,62 +333,56 @@ describe('GDP.ADVANCED.model.Job', function() {
...
+			]);
 		});
 
-		it('Expects a callWFS to be issued with the appropriate typename, propertyname and cql_filter', function() {
-			jobModel.getSelectedFeatureIds();
-			expect(GDP.OGC.WFS.callWFS).toHaveBeenCalled();
-			expect(GDP.OGC.WFS.callWFS.calls[0].args[0].typename).toEqual('derivative:CONUS_States');
-			expect(GDP.OGC.WFS.callWFS.calls[0].args[0].propertyname).toEqual('FeatureAttribute');
-			expect(GDP.OGC.WFS.callWFS.calls[0].args[0].cql_filter).toEqual('Fake filter');
+		it('Expects that is aoiAttributeValues is the empty array that getSelectFeatureIds returns the empty array', function() {
```
cschroed-usgs: Do you mean "Expects that if..."?

src/main/webapp/js/advanced/models/JobModel.js
cschroed-usgs: Oh wait, I get it -- e.ids is an array and you want the `result` array to be flat. The way you do it makes sense since there is no [flatmap operation](http://martinfowler.com/articles/collection-pipeline/flat-map.html) in JS or in underscore. You could also do this by stringing a _.map and a _.flatten call together, but I don't see a clear advantage of doing that vs your current approach.

src/main/webapp/WEB-INF/jsp/config.jsp
mbucknell: I tried that but the application throws an error and my IDE indicates that there is no method with the signature.

src/main/webapp/js/advanced/models/JobModel.js
mbucknell: That's one reason. The other is clarity of intent. I think sometimes as programs we try to put too many things in one statement which can make it harder for a maintainer to figure out what's going.

src/main/webapp/js/advanced/templates/spatial.html
mbucknell: It may be put I'm going to let Blodgett make this call. I think I will change 'Feature' in the above sentence to Area of Interest though.

src/main/webapp/js/advanced/views/SpatialView.js
mbucknell: I'm going to leave it as a string. That way all config values are Strings.

mbucknell:
I already started the next GDP JIRA which was a major naming refactor. I made some of the requested changes/clarifications in there. I will go ahead and merge this as is.

src/main/webapp/WEB-INF/jsp/config.jsp
cschroed-usgs: Sorry, I screwed up. The one-liner should say 'getProperty', not 'get':
```
String maxPolygonsToShowAttributeValues = props.getProperty("gdp.application.maxPolygonsToShowAttributeValues", "1000");
```
Does that work?

mbucknell: yes ... I'll update it the next time I do a pull request for GDP.

src/main/webapp/js/advanced/models/JobModel.js
cschroed-usgs: I agree, shorter is not always better. On the other hand, if you take that argument to an extreme (and I know you're not doing that) then we're all programming in assembly :) Finding that sweet balance of brevity and clarity is a context-dependent art. I think that the features I mentioned above provide novel tools that would lend brevity and clarity to a lot of our app development use cases. It turns out that there's a name for this pattern: Collection Pipelining. If you're interested in reading more, Fowler has a good article on it. http://martinfowler.com/articles/collection-pipeline/#Parallelism/

src/main/webapp/WEB-INF/jsp/config.jsp
cschroed-usgs: Sweet!

PR #62, opened 2015-05-28
Gdp 933

mbucknell:
Initial implementation of the landing page showing all data sets retrieved from sciencebase. Each dataset is rendered as a clickable tile. Clicking a tile brings up more information about a data set, including data sources and available algorithms. Clicking Process Data will route that user to the gdp's process_client url with the identifier in the url. At the moment, the routing works,b ut rendering for the page has not been added to the application.

DataSetModel is where information about each model is stored. DataSetSelectionView is the main page view, which then creates a DataSetTileView for each dataset. DataSetDialogView creates the modal dialog content.

### DIFF
--- a/src/main/webapp/js/landing/init.js
+++ b/src/main/webapp/js/landing/init.js
@@ -12,7 +12,8 @@ $(document).ready(function() {
 
 	var TEMPLATES = [
 		'datasource_select',
-		'data_set_tile'
+		'data_set_tile',
+		'data_set_dialog'
 	];
 
 	GDP.LANDING.templates = GDP.util.templateLoader('js/landing/templates/');
@@ -34,12 +35,21 @@ $(document).ready(function() {
 			}
 		}));
 
+	var loadAlgorithms = $.ajax('algorithms', {
+		success : function(data) {
+			GDP.algorithms = new Backbone.Model($.parseJSON(data));
+		},
+		error : function(jqXHR, textStatus) {
+			console.log('Can\'t load algorithms ' + textStatus);
+		}
+	})
+
 	// I need to load up my config model since one of the views I load depends on it
 	// Load up the process collection based on incoming model definitions from the config object
 
 	var loadTemplates = GDP.LANDING.templates.loadTemplates(TEMPLATES);
 
-	$.when(loadTemplates, loadConfigModel).always(function () {
+	$.when(loadTemplates, loadConfigModel, loadAlgorithms).always(function () {
 		GDP.LANDING.templates.registerHelpers();
 
 		var dataSets = new GDP.LANDING.models.DataSetCollection();
--- a/src/main/webapp/js/landing/init.js
+++ b/src/main/webapp/js/landing/init.js
@@ -40,6 +40,8 @@ $(document).ready(function() {
 	var loadTemplates = GDP.LANDING.templates.loadTemplates(TEMPLATES);
 
 	$.when(loadTemplates, loadConfigModel).always(function () {
+		GDP.LANDING.templates.registerHelpers();
+
 		var dataSets = new GDP.LANDING.models.DataSetCollection();
 		GDP.LANDING.router = new GDP.LANDING.controller.LandingRouter({
 			dataSets : dataSets
--- a/src/main/webapp/js/landing/init.js
+++ b/src/main/webapp/js/landing/init.js
@@ -1,0 +1,42 @@
+/*jsline browser: true*/
+/*global log4javascript*/
+/*global $*/
+/*global Backbone*/
+
+var GDP = GDP || {};
+
+GDP.LANDING = GDP.LANDING || {};
+
+$(document).ready(function() {
+	"use strict";
+
+	var TEMPLATES = [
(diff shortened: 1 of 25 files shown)


src/main/webapp/js/landing/init.js
```diff
@@ -0,0 +1,63 @@
...
+			console.log('Can\'t load algorithms ' + textStatus);
+		}
+	})
+
+	// I need to load up my config model since one of the views I load depends on it
+	// Load up the process collection based on incoming model definitions from the config object
+
+	var loadTemplates = GDP.LANDING.templates.loadTemplates(TEMPLATES);
+
+	$.when(loadTemplates, loadConfigModel, loadAlgorithms).always(function () {
```
thongsav-usgs: @jiwalker-usgs: you might want to consider this pattern for that CCH problem with loading a set of SLDs.

jiwalker-usgs: Yeah, we decided to put that off for now, but I'll keep this in mind when we come back to that.

src/main/webapp/js/landing/models/DataSetModel.js
```diff
@@ -0,0 +1,173 @@
...
+var GDP = GDP || {};
+
+GDP.LANDING = GDP.LANDING || {};
+
+GDP.LANDING.models = GDP.LANDING.models || {};
+
+(function() {
+	"use strict";
+
+	GDP.LANDING.models.DataSetModel = Backbone.Model.extend({
```
thongsav-usgs: Do you think it'd be beneficial to do a full blown Backbone model hierarchy here, where the DataSetModel consists of 4 submodels (DataSourcesModel, ContactInfoModel, etc...). If the results of the getters are always the same, you may be able to get away with performing the logic once in the init/constructor of the submodels.

src/main/webapp/js/landing/templates/data_set_dialog.html
```diff
@@ -0,0 +1,64 @@
...
+		<dl>
+			<dt>Distribution Info</dt>
+			<dd>
+			{{#if distributionInfo}}
+			{{distributionInfo.name}} - <a href="{{distributionInfo.url}}">{{distributionInfo.url}}</a>
+			{{else}}Not Available
+			{{/if}}
+			</dd>
+		</dl>
+		<div id="data-sources-div">
```
thongsav-usgs: I'm still not a fan of using static IDs in templates.

@isuftin, this is a good instance of where I don't like to use ids. This is for a dialog box. If you throw up multiple dialogue windows and write to this ID, all dialogue boxes will have their content changed since they are sharing the same div.

src/main/webapp/js/landing/models/DataSetModel.js
mbucknell: I want to preserve the isoMetadata object as it is received from the CSW service, so having a model hierarchy doesn't make sense to me.

mbucknell:
This is a place where I probably should not use it. Although right now there is only one instance of this template at a time I can see that changing. I'll change this one to a class.

src/main/webapp/js/landing/views/DataSetDialogView.js
```diff
@@ -0,0 +1,101 @@
...
+		removeDialog : function() {
+			this.$el.modal('hide');
+			this.remove();
+		},
+
+		/*
+		 * Go to the process_client using the dataset identifier in the url.
+		 */
+		goToProcessClient : function() {
+			window.location.assign('process_client/' + this.model.get('csw').identifier);
```
thongsav-usgs: Probably doesn't matter much, but is there a router available to this view to go through?

mbucknell: No .. it's actually a separate page within the GDP app. So it will be in that page's router (although it's not there at the moment), but it won't work since that router is not in this page's scope.

src/main/webapp/js/ogc/csw.js
```diff
@@ -1,32 +1,95 @@
...
+			}
+		});
+		return deferred.promise();
 	};
+
+	this.requestGetRecordById = function(args) {
+		var outputSchema = args.outputSchema ? args.outputSchema : 'http://www.opengis.net/cat/csw/2.0.2';
+		var id = args.id;
+		var cswGetRecFormat = new OpenLayers.Format.CSWGetRecords.v2_0_2();
+		var deferred = $.Deferred();
```
thongsav-usgs: :+1: This makes me want to go back and untangle crazy ajax chains I've created in other projects.

mbucknell: Me too! I have been making some bug fixes for WQP and the use of callbacks is making me cringe.

thongsav-usgs:
:+1:

src/main/webapp/js/ogc/csw.js
cschroed-usgs: @thongsav-usgs: Deferreds are awesome. You will love them. I "promise". DON'T HURT ME FOR MY PUNZ :D

PR #63, opened 2015-06-01
Gdp 945

mbucknell:
Need to remove inputs that are defined in config.jsp for the selected algorithm. Updated the tests to reflect this new requirement.


thongsav-usgs:
:+1:

PR #67, opened 2015-06-02
Updated DataSetModel to initialize model attributes directly from the…

mbucknell:
… metadata record. Decided to get the data as ISO-19115 so that in the future other catalog services can be used without modifying the code. Then it made sense to directly parse the metadata record into attributes that can be used by the views and templates.

Updated the tests to reflect the new code.


mbucknell:
@thongsav-usgs, I took your earlier suggestion about setting attributes in DataSetModel at initialization. Partially driven by a change/clarification of application requirements. Rather than creating a hierarchy of models though, I decided to use the Backbone model's parse function. This makes sense to me. If we were using a typically ajax call this is what we would use.

mbucknell:
I've added the rest of the changes for GDP-949. This works adds a sidebar to the landing page and puts a single text field to be used to filter the datasets using the title and abstract fields.

mbucknell:
GDP-949 and GDP-950's changes are both in pull request 68.

PR #68, opened 2015-06-04
Gdp 950

mbucknell:
This combines GDP 949 and GDP 950. The functionality added was filtering by text and by available algorithm. I also refactored DataSetModel to use the parse function when creating the model. The metadata record is passed directly to the constructor.

Tests have been updated to test new functionality.


thongsav-usgs:
:+1:

PR #69, opened 2015-06-08
Gdp 948

mbucknell:
This pull request partially implements GDP-948. I have not yet added the meta data tile to the hub page. This code implements passing the dataset id in the url and then having the available algorithms limited to what is available and have the data sources selected from a pick list.

Since we want urls without hashes, I activated pushState routing. This required loading all static using the full URL, so I added a JNDI property, gdp.base.url.


mbucknell:
I'm closing this because I have a new one https://github.com/USGS-CIDA/gdp_web/pull/70 which completely implement GDP-948.

PR #70, opened 2015-06-08
Gdp 948 metadata

mbucknell:
This pull request partially implements GDP-948. I have not yet added the meta data tile to the hub page. This code implements passing the dataset id in the url and then having the available algorithms limited to what is available and have the data sources selected from a pick list.

Since we want urls without hashes, I activated pushState routing. This required loading all static using the full URL, so I added a JNDI property, gdp.base.url.

I have also refactored the templates to put them at the root directory rather than within the js namespaced directories. This allows templates to be more easily shared.


src/main/webapp/WEB-INF/jsp/config.jsp
```diff
@@ -22,6 +22,7 @@
 	String maxPolygonsToShowAttributeValues = props.get("gdp.application.maxPolygonsToShowAttributeValues");
 	maxPolygonsToShowAttributeValues = (maxPolygonsToShowAttributeValues == null) ? "1000" : maxPolygonsToShowAttributeValues;
 	String geoserverEndpoint = props.get("gdp.geoserver.endpoint");
+	String baseUrl = props.getProperty("gdp.base.url");
```
isuftin: Might want to do this here:

    if (!baseUrl.endsWith("/")) { baseUrl += "/"; }

src/main/webapp/WEB-INF/jsp/head.jsp
```diff
@@ -16,6 +16,7 @@
 %>
 <%
 Boolean development = Boolean.parseBoolean(props.getProperty("gdp.development"));
+String baseUrl = props.getProperty("gdp.base.url");
```
isuftin: Same comment as above (or you can probably move this functionality into an included base.jsp which gets inline included into here, though I should verify that works properly before suggesting it)

mbucknell: I'm wondering if I need it in both places. I did notice development is also defined in both.

src/main/webapp/WEB-INF/jsp/head.jsp
```diff
@@ -31,11 +32,11 @@ String resourceSuffix = development ? "" : "-" + versionProject + "-min";
...
-			<jsp:param name="relPath" value="" />
+			<jsp:param name="relPath" value="<%= baseUrl %>" />
 			<jsp:param name="shortName" value="USGS Geo Data Portal" />
 			<jsp:param name="title" value="USGS Geo Data Portal" />
 			<jsp:param name="description" value="" />
-			<jsp:param name="author" value="Ivan Suftin" />
+			<jsp:param name="author" value="" />
 			<jsp:param name="keywords" value="" />
 			<jsp:param name="publisher" value="" />
 			<jsp:param name="revisedDate" value="" />
```
isuftin: I wonder if we can set the maven build date as a property (in the same way we inject js versions, etc)  and inject into here.

Just a thought. I don't think we currently do that at all anywhere.

src/main/webapp/WEB-INF/jsp/head.jsp
isuftin: If head is an included JSP, that can probably be passed into the jsp via a parameter

src/main/webapp/index.jsp
```diff
@@ -54,7 +54,7 @@
 		<script type="text/javascript" src="js/util/mapUtils<%= resourceSuffix %>.js"></script>
 		<script type="text/javascript" src="js/ogc/csw<%= resourceSuffix %>.js"></script>
 		<script type="text/javascript" src="js/models/Config<%= resourceSuffix %>.js"></script>
-		<script type="text/javascript" src="js/landing/models/DataSetModel<%= resourceSuffix %>.js"></script>	
+		<script type="text/javascript" src="js/models/DataSetModel<%= resourceSuffix %>.js"></script>
```
isuftin: One thing we've started doing on CCH is adding the version number to our JS files so there's no JS caching post-release for users that have already visited the site. 

Something like [this](https://github.com/USGS-CIDA/coastal-hazards/blob/master/coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/ui/front/index.jsp#L34)

Might be useful for other projects as well.

src/main/webapp/js/landing/views/DataSetDialogView.js
```diff
@@ -65,7 +65,7 @@ GDP.LANDING.views = GDP.LANDING.views || {};
 		 * Go to the process_client using the dataset identifier in the url.
 		 */
 		goToProcessClient : function() {
-			window.location.assign('process_client/' + this.model.get('csw').identifier);
+			window.location.assign('client/catalog/gdp/dataset/' + this.model.get('identifier') + '');
```
isuftin: Why the empty quotes at the end? I'm assuming this was to create a string if "this.model.get('identifier')" was undefined or null, but now you have a string before that so you shouldn't need the empty quotes.

src/main/webapp/index.jsp
mbucknell: Yep ... you clued me into that before and I'm using it here. If you look in head.jsp, that where resourceSuffix is defined using the version.

src/main/webapp/js/models/DataSetModel.js
```diff
@@ -23,6 +21,9 @@ GDP.LANDING.models = GDP.LANDING.models || {};
 		 * @param {Object} options
 		 */
 		parse : function(metadata, options) {
+			if (_.isEmpty(metadata)) {
+				return {};
+			}
 			var result = {};
```
isuftin: OCD...

    var result = {};

    if (_.isEmpty(metadata)) {
        return result;
    }

:D

src/main/webapp/WEB-INF/jsp/head.jsp
mbucknell: Has to be in both because the other is config.jsp which is read in via the javascript.

isuftin: Sure but if config.jsp is in-line included, that should seamlessly become a part of head.jsp (if config.jsp is in-line included in head.jsp). In this way, the JS in config.jsp is also a part of your config.jsp and baseUrl is only needed once. 

Check this out to see what I mean: http://www.jguru.com/faq/view.jsp?EID=13517

I'm not saying the refactor should be part of this PR, but keep it in mind if you're going to refactor a bit in the future.

src/main/webapp/js/process_client/init.js
```diff
@@ -37,11 +45,16 @@ $(document).ready(function() {
...
-	$.when(loadTemplates, loadPartials, loadConfigModel).always(function () {
-		initializeLogging({
-			LOG4JS_LOG_THRESHOLD: GDP.config.get('application').development === 'true' ? 'debug' : 'info'
-		});
+	var loadAlgorithms = $.ajax(GDP.BASE_URL + 'algorithms', {
+		success : function(data) {
+			GDP.algorithms = new Backbone.Model($.parseJSON(data));
+		},
+		error : function(jqXHR, textStatus) {
+			console.log('Can\'t load algorithms ' + textStatus);
```
isuftin: Use Log4JS here

src/main/webapp/js/process_client/init.js
```diff
@@ -15,16 +15,24 @@ $(document).ready(function() {
...
 	var PARTIALS = [];
 
-	GDP.PROCESS_CLIENT.templates = GDP.util.templateLoader('js/process_client/templates/');
+	GDP.PROCESS_CLIENT.templates = GDP.util.templateLoader(GDP.BASE_URL + 'templates/');
 
-	var loadConfigModel = $.when($.ajax('config', {
+	var loadConfigModel = $.when($.ajax(GDP.BASE_URL + 'config', {
 			success: function (data) {
 				GDP.config = new GDP.model.Config(data);
+				var applicationConfig = GDP.config.get('application');
```
isuftin: I'd suggest initializing logging as the first order of business in this file outside of this AJAX call. You want Log4JS available as soon as possible and then use it going forward instead of console.log

src/main/webapp/js/process_client/init.js
```diff
@@ -37,11 +45,16 @@ $(document).ready(function() {
...
+	var loadAlgorithms = $.ajax(GDP.BASE_URL + 'algorithms', {
+		success : function(data) {
+			GDP.algorithms = new Backbone.Model($.parseJSON(data));
+		},
+		error : function(jqXHR, textStatus) {
+			console.log('Can\'t load algorithms ' + textStatus);
+		}
+	});
 
+	$.when(loadTemplates, loadPartials, loadConfigModel, loadAlgorithms).always(function () {
```
isuftin: I really need to start organizing my callbacks in this way. The spaghetti code I write makes sense this month but next month, I guarantee it won't and this absolutely helps. I'm curious as to why you wrap loadTemplates, loadPartials and loadConfigModel in a $.when() call but not loadAlgorithms?

src/main/webapp/js/process_client/models/JobModel.js
```diff
@@ -64,6 +65,47 @@ var GDP = GDP || {};
...
 
 		/*
+		 *
+		 * @param {String} datasetId - Id of dataset to retrieve from sciencebase
+		 * @returns {jquery.Promise} - This promise is resolve if the data set model does not need to be updated or if it is
+		 * successfully updated. The promise is rejected if updating the datset failed. The dataSetModel property is also cleared.
+		 */
+		updateDataSetModel : function(datasetId) {
+			var dataSetModel = this.get('dataSetModel');
+			var deferred = $.Deferred();
```
isuftin: I like this pattern. My only argument against something like this is that the calling function needs to have intimate knowledge of what's happening inside of this function. Is that not the case?

src/main/webapp/js/process_client/views/DataDetailsView.js
```diff
@@ -35,25 +40,38 @@ var GDP = GDP || {};
 	}()),
 	'wps' : null,
 	'initialize': function(options) {
+		var self = this;
+		var initArguments = arguments;
+
 	    this.wps = options.wps;
+		this.routePrefix = options.datasetId ? 'catalog/gdp/dataset/' + options.datasetId : 'advanced';
```
isuftin: weird tabbing

src/main/webapp/less/datadetail.less
```diff
@@ -19,4 +19,3 @@
 		.make-sm-column(4);
 	}
 }
-
```
isuftin: ?

src/test/js/specs/JobModel_spec.js
```diff
@@ -396,4 +396,77 @@ describe('GDP.PROCESS_CLIENT.model.Job', function() {
...
+		});
+
+		it('Expects that if the datasetId is the same as the current dataSetModel that the model is not updated', function() {
+			var promise = jobModel.updateDataSetModel('ID1');
+			expect(GDP.cswClient.requestGetRecordById).not.toHaveBeenCalled();
+			expect(promise.state()).toBe('resolved');
+			expect(jobModel.get('dataSetModel').attributes.identifier).toBe('ID1');
+		});
+
+		it('Expect that if a new datasetId is specified, the record is retrieved and updated if the retrieval is successful', function() {
```
isuftin: Expects

;)

src/main/webapp/js/process_client/init.js
mbucknell: That's the way the code was ... it's not necessary because the $.ajax call returns a promise. I'll remove them.

src/main/webapp/js/process_client/models/JobModel.js
mbucknell: I don't think so. This one is fairly straightforward, a resolved promise means the data set model has been updated, cleared if no datasetId is set, retrieved if it is. This does cache the response so that another call is not made if the datasetId is the same as the model's identifier property. The comment really is all the user needs to know.

src/main/webapp/js/process_client/init.js
isuftin: Ah I thought cleverness was happening :) 

I was playing with passing ajax xhr objects around yesterday and the part I couldn't (and maybe still can't) get my head wrapped around is what triggers the actual ajax call and then I realized it's triggered as soon as I create the ajax object even before I pass it back. 

I feel like there's a bit of loss of control there but maybe that's just in my own head.

src/main/webapp/js/process_client/models/JobModel.js
isuftin: So if everything is done within this function, what's the purpose of returning the deferred? Is it for testing purposes?

I am of the mindset that *all* functions should ideally return something, but often I'm stuck trying to figure out what to return and if it's meaningful.

More on that mindset: http://stackoverflow.com/questions/20915450/why-javascript-functions-always-return-a-value

src/main/webapp/js/process_client/init.js
mbucknell: Can't use the logger yet because you don't know that it has been initialized.

src/main/webapp/js/process_client/init.js
mbucknell: I'm not familar with the log4js logger. The InitializeLogging can't be done until the config has been loaded. Shouldn't this be done before creating the logger?

src/main/webapp/js/process_client/init.js
isuftin: Right, but if you init it outside of the ajax call above, it should be initialized by the time this ajax call even begins

src/main/webapp/js/process_client/init.js
isuftin: You need config to be loaded because you want to know if we're development or prod. We already know this when the JSP that loads this JS is loaded so we can just get it from there.

src/main/webapp/js/process_client/models/JobModel.js
mbucknell: Returning the promise (which can't be resolved or rejected) allows the caller to perform processing once the loading is complete. In this code for instance HubView calls it (line 66) and will display an alert if the load failed. Once the load has completed (successfully or not), it finishes rendering the page.

isuftin: Ok so basically this is a way to do core functionality here but also allow calling functions to attach riders if they wish?

mbucknell: Just like our old friend, the callback!

isuftin: Lovin' it. I usually pass in error and success callbacks to functions that perform ajax calls and then run through the array of each. ( [example](https://github.com/isuftin/coastal-hazards/blob/master/coastal-hazards-portal/src/main/webapp/js/cch/objects/Session.js#L90-L125) )

mbucknell:
Updated everything but did not refactor to include config.jsp in head.jsp. This threw an error and didn't want to spend the time right now to untangle it. Ok to merge?

isuftin:
Ayyy

PR #72, opened 2015-06-11
Gdp 929

mbucknell:
Added a layer representing a dataSet/dataSources extent on the map on the spatial page and the map on the hub page. If a dataSource has not yet been specified, then show the extent of the dataset. If it has, do a getCapabilities call to the catalogwms service to determine if a extent layer exists. If not, just continue to use the dataset extent.

Note that if the user comes from the "advanced" route, no extent layer is shown.

### DIFF
--- a/src/main/webapp/WEB-INF/jsp/config.jsp
+++ b/src/main/webapp/WEB-INF/jsp/config.jsp
@@ -35,7 +35,7 @@
 			"utilityWps" : "<%= baseUrl %>utilityWps",
 			"processWps" : "<%= baseUrl %>processWps",
 			"csw" : "<%= baseUrl %>csw",
-			"catalogwms" : "<%= baseUrl %>catalogwms"
+			"catalogWms" : "<%= baseUrl %>catalogwms"
 		},
 		"serviceEndpoints" : {
 			"geoserver" : "<%= geoserverEndpoint %>"
--- a/src/main/webapp/WEB-INF/jsp/config.jsp
+++ b/src/main/webapp/WEB-INF/jsp/config.jsp
@@ -34,7 +34,8 @@
 			"geoserver" : "<%= baseUrl %>geoserver",
 			"utilityWps" : "<%= baseUrl %>utilityWps",
 			"processWps" : "<%= baseUrl %>processWps",
-			"csw" : "<%= baseUrl %>csw"
+			"csw" : "<%= baseUrl %>csw",
+			"catalogwms" : "<%= baseUrl %>catalogwms"
 		},
 		"serviceEndpoints" : {
 			"geoserver" : "<%= geoserverEndpoint %>"
--- a/src/main/webapp/js/process_client/views/SpatialView.js
+++ b/src/main/webapp/js/process_client/views/SpatialView.js
@@ -668,22 +668,25 @@ GDP.PROCESS_CLIENT.view = GDP.PROCESS_CLIENT.view || {};
 		},
 
 		_addDatasetBoundingBoxLayer : function() {
+			var self = this;
 			var dataSetModel = this.model.get('dataSetModel');
+			var dataSourceUrl = this.model.get('dataSourceUrl');
 			var bounds;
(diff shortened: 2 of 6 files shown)


src/main/webapp/js/process_client/views/SpatialView.js
```diff
@@ -661,6 +667,29 @@ GDP.PROCESS_CLIENT.view = GDP.PROCESS_CLIENT.view || {};
...
+			var dataSetModel = this.model.get('dataSetModel');
+			var dataSourceUrl = this.model.get('dataSourceUrl');
+			var bounds;
+
+			if (dataSetModel.has('identifier') && dataSetModel.has('bounds')) {
+				bounds = dataSetModel.get('bounds');
+				if (dataSourceUrl) {
+					GDP.util.mapUtils.createDataSourceExtentLayer(bounds, dataSetModel.get('identifier'), dataSourceUrl).done(function(layer) {
+						self.boundsLayer = layer;
+						self.map.addLayer(self.boundsLayer);
```
isuftin: You may want to do a pre-check to ensure that this layer does not already exist in the map

src/main/webapp/js/util/mapUtils.js
```diff
@@ -141,6 +141,96 @@ GDP.util.mapUtils = (function() {
...
+			params : {
+				request : 'GetCapabilities',
+				service : 'WMS',
+				version : '1.3.0'
+			},
+			success : function(response) {
+				var respObj = format.read(response.responseXML);
+				if (_.has(respObj, 'error')) {
+					GDP.debug('Unsuccessfully retrieved wms capabilities');
+					layer = that.createDataSetExtentLayer(boundingBox);
```
isuftin: It might be cleaner to set this at 173 and then you don't have to have all of these if statements. You'd just need the one success statement (starting at line 197) and then your onError is just the single line: deferred.resolve(layer);

src/main/webapp/WEB-INF/jsp/config.jsp
```diff
@@ -34,7 +34,8 @@
 			"geoserver" : "<%= baseUrl %>geoserver",
 			"utilityWps" : "<%= baseUrl %>utilityWps",
 			"processWps" : "<%= baseUrl %>processWps",
-			"csw" : "<%= baseUrl %>csw"
+			"csw" : "<%= baseUrl %>csw",
+			"catalogwms" : "<%= baseUrl %>catalogwms"
```
isuftin: Moving away from camelcasing like above?

src/main/webapp/js/util/mapUtils.js
mbucknell: I didn't want to create the layer object just to throw it away if everything succeed. That seems wrong since itcreates a "new" object.

src/main/webapp/WEB-INF/jsp/config.jsp
mbucknell: I'm not a big fan of camel casing urls.

isuftin: Sure but what about the key name?

src/main/webapp/js/util/mapUtils.js
isuftin: What are we programming for, a calculator watch?? :D Don't worry about it :D

src/main/webapp/WEB-INF/jsp/config.jsp
mbucknell: Ok ... I'll change it.

src/main/webapp/js/util/mapUtils.js
mbucknell: I don't think it makes it easier to understand what's going on though. The rules for creating this layer are likely to change as we go along. I'd like to leave it as is.

isuftin: Ok

isuftin:
:+1:

Issue #75, opened 2015-06-22
Tips on deployment?


rsignell-usgs:
If I wanted to deploy this locally for testing/experimentation, what would the steps be?

cschroed-usgs:
Hi Rich!

Thanks for your interest in running gdp_web. I updated the app's README.md with some instructions. I'm not familiar with your technical background. Apologies in advance if I omitted something important or stated the obvious. How do the instructions work for you?

cschroed-usgs:
Rich,

It turns out my instructions alone are inadequate. There are some additional configuration steps to perform. @mbucknell will likely address these requirements.

rsignell-usgs:
@cschroed-usgs , okay, thanks for the heads up.   I was just about to try it, but was wondering whether certain ports had to be opened.    We already have java 7 and tomcat 7 running.

dblodgett-usgs:
Hi Rich, You will need a number of entries in your context.xml. We will add what these are to the readme when time allows. Mary is off on another project and will get to this next time she works on this one. You aren't in a huge rush to get to this are you?

rsignell-usgs:
No, not in a rush.

mbucknell:
@rsignell-usgs, I have updated the README.md to include the required JNDI variables.

PR #77, opened 2015-06-22
Update README.md

cschroed-usgs:
Addressing #75.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,19 @@
 # gdp_web
 Web UI for Geo Data Portal
+
+## To run ##
+install git, java 7, and maven 3, and tomcat 7
+
+clone this repo
+
+open a terminal, change to the directory where you cloned the repository
+
+build a .war artifact by running the command:
+```
+mvn clean package
+```
+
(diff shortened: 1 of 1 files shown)


jiwalker-usgs:
Is there any configuration beyond this that is necessary?

mbucknell:
There are a whole set of JNDI properties needed in order to run this. The app is pretty tightly coupled to the gdp wps processing server.

cschroed-usgs:
@jiwalker-usgs , maybe. Thanks for bringing it up. @mbucknell, @isuftin , correct me if I'm wrong, but I think we were keeping an context.xml in a google doc somewhere instead of checking an exampleContext.xml in to the repo like we do on most other projects. Our collaborator's interest in running the project highlights the weakness of the google drive approach. We prefer reproducible builds.

mbucknell:
The configuration for this app is actually controlled by a set of files in stash that use ansible. We are no longer maintaining that google doc.

cschroed-usgs:
That's good to hear. I'll defer to you on whether it's better to include that file from the stash repo in the github repo, or send the file to Rich directly.

PR #99, opened 2015-08-28
Gdp 996

mamcderm-usgs:
@mbucknell - It seemed like all the error handling was in place.  I did fix one javascript problem and changed the messaging on the landing page to be consistent with NWC.  For the most part I left the other views the same as they already produced an error message.  This seemed adequate to me but let me know if you think any of that should be modified.

### DIFF
--- a/src/main/webapp/js/process_client/views/ProcessView.js
+++ b/src/main/webapp/js/process_client/views/ProcessView.js
@@ -28,12 +28,15 @@ GDP.PROCESS_CLIENT.view = GDP.PROCESS_CLIENT.view || {};
 
 			this.algorithmTemplate = options.algorithmTemplate;
 			this.routePrefix = options.datasetId ? 'catalog/gdp/dataset/' + options.datasetId  : 'advanced';
-
+			
 			this.model.updateDataSetModel(options.datasetId).always(function() {
 				GDP.util.BaseView.prototype.initialize.apply(self, initArguments);
 				self.listenTo(self.model, 'change:algorithmId', self.displayAlgorithmDescription);
 			}).fail(function() {
-			    window.alert('Unable to load requested dataset ' + options.datasetId);
+				self.alertView = new GDP.util.AlertView({
+					el : '#messages-div'
+				});
+				self.alertView.show('alert-danger', 'Unable to load information about the dataset, ' + options.datasetId);
 			});
 		},
 
--- a/src/main/webapp/js/process_client/views/ProcessView.js
+++ b/src/main/webapp/js/process_client/views/ProcessView.js
@@ -28,12 +28,12 @@ GDP.PROCESS_CLIENT.view = GDP.PROCESS_CLIENT.view || {};
 
 			this.algorithmTemplate = options.algorithmTemplate;
 			this.routePrefix = options.datasetId ? 'catalog/gdp/dataset/' + options.datasetId  : 'advanced';
-                        
+
 			this.model.updateDataSetModel(options.datasetId).always(function() {
 				GDP.util.BaseView.prototype.initialize.apply(self, initArguments);
 				self.listenTo(self.model, 'change:algorithmId', self.displayAlgorithmDescription);
 			}).fail(function() {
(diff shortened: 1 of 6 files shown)


src/main/webapp/js/process_client/views/ProcessView.js
```diff
@@ -28,12 +28,12 @@ GDP.PROCESS_CLIENT.view = GDP.PROCESS_CLIENT.view || {};
...
 			this.algorithmTemplate = options.algorithmTemplate;
 			this.routePrefix = options.datasetId ? 'catalog/gdp/dataset/' + options.datasetId  : 'advanced';
-
+                        
 			this.model.updateDataSetModel(options.datasetId).always(function() {
 				GDP.util.BaseView.prototype.initialize.apply(self, initArguments);
 				self.listenTo(self.model, 'change:algorithmId', self.displayAlgorithmDescription);
 			}).fail(function() {
-				window.alert('Unable to load requested dataset ' + options.datasetId);
+                            window.alert('Unable to load requested dataset ' + options.datasetId);
```
mbucknell: The other tile view use alertView rather than the window.alert. Can we switch to using that one here?

mbucknell:
See the one comment. I'm also seeing some formatting issues. Can you make sure that your netbeans is using tabs and not spaces?

src/main/webapp/js/process_client/views/ProcessView.js
mamcderm-usgs: I started going down that road and it seemed to use window.alert for some timing reason that I was leery about messing with.  But I can do that.

mamcderm-usgs:
I was trying to fix the formatting issues and I can't see the problem.  I think I have netbeans set to use tabs but maybe I am missing a setting somewhere.  It is set to use "Tabs and Indents" in Tools -> Options -> Editor -> Formatting.  Is there another place?

src/main/webapp/js/process_client/views/ProcessView.js
mbucknell: I don't have any clear memory of this, but it is strange that this is the only one that uses the window.alert. Let's see if it can be switched. If you aren't having any look, let's skip it for now.

mbucknell:
what version of netbeans are you running? I don't have that in my Tools drop down. Instead, in Preferences, in the Editor tab and then select Formatting. Select JavaScript as the language, category as Tabs and Indents and then the Expand Tabs to Spaces should not be checked. I also have the JSHint plugin installed which will help with finding javascript code issues.

src/main/webapp/js/process_client/views/ProcessView.js
mamcderm-usgs: Actually DataDetailsView uses window.alert also.  But ProcessView doesn't doesn't render until after initialize because of this comment in the code:

// Hiding the rendered template until the selected algorithm description is shown so that
// there isn't flashing when an algorithm has already been selected

I wasn't exactly sure what the effect would be if I use alertView because I would need to call initialize.apply.  I can give it a try though.

mamcderm-usgs:
I'm on 8.0.2.  I don't see any Preferences menu option.

src/main/webapp/js/process_client/views/ProcessView.js
mbucknell: I wouldn't do that. You could wait to handle the fail until the view has been rendered. This would mean putting the promise onto the scope of the view, so something like this: this.updateDataModelPromise = this.model.updateDataSetModel .....

mamcderm-usgs:
arggg - Editor is still converting tabs to spaces even though I don't have it checked...I even closed and reopened netbeans.  Lots of stuff online about the problem but no obvious solution.

mamcderm-usgs:
@mbucknell - I think I have this in better shape.  Do you want to take a look and see if you have any other suggestions.

mbucknell:
Looks good to go!

mamcderm-usgs:
@mbucknell - It doesn't appear I have permissions to merge on this repos.  Do you know how I would make that request?

mbucknell:
You should be able to merge now.

PR #109, opened 2015-09-14
Gdp 985

mbucknell:
Refactored the DataDetailsView to move the fetching of the data variables and min/max Dates into DataSourceModels. The models were also reworked so that the model represented the data received. The view is then responsible for putting the data into a form that can be used by SelectMenuView.

The views no longer depend on wps as this is used in the models to fetch the data. I made this a namespaced global, GDP.wpsClient  similar to what we do with GDP.cswClient.

As part of this refactor, made the data details inputs disabled while the data source model is being fetched.

### DIFF
--- a/src/main/webapp/js/process_client/models/DataSourceModels.js
+++ b/src/main/webapp/js/process_client/models/DataSourceModels.js
@@ -1,5 +1,6 @@
 /*jslint browser: true*/
 /*global Backbone*/
+/*global _*/
 var GDP = GDP || {};
 (function(){
     "use strict";
@@ -7,18 +8,168 @@ var GDP = GDP || {};
 
     GDP.PROCESS_CLIENT.model = GDP.PROCESS_CLIENT.model || {};
 
+	var VARIABLE_WPS_PROCESS_ID = 'gov.usgs.cida.gdp.wps.algorithm.discovery.ListOpendapGrids';
+	var DATE_RANGE_WPS_PROCESS_ID = 'gov.usgs.cida.gdp.wps.algorithm.discovery.GetGridTimeRange';
+
     var DataSourceVariable = Backbone.Model.extend({
-	defaults:{
-	    text: '',
-	    value: null,
-	    selected: false
-	}
-    });
+		defaults:{
+			name : '',
+			description : '',
+			unitsstring : ''
+		}
+	});
 
     var DataSourceVariables = Backbone.Collection.extend({
-	model: DataSourceVariable
+		model: DataSourceVariable,
+
+		fetch : function(options) {
+			var self = this;
+			var deferred = $.Deferred();
+			var wpsInputs = {
+				"catalog-url": [options.dataSourceUrl],
+				"allow-cached-response": [options.allow_cached]
+			};
+			var wpsOutput = ["result_as_json"];
+
+			GDP.wpsClient.sendWpsExecuteRequest(
+				GDP.config.get('application').endpoints.utilityWps + '/WebProcessingService',
+				VARIABLE_WPS_PROCESS_ID,
+				wpsInputs,
+				wpsOutput,
+				false,
+				null,
+				true,
+				'json',
+				'application/json'
+			).done(function(response) {
+				var dataCollection;
+				if (response && _.has(response, 'datatypecollection') && _.has(response.datatypecollection, 'types')) {
+					if (_.has(response.datatypecollection.types, 'length')) {
+						dataCollection = response.datatypecollection.types;
+					}
+					else {
+						dataCollection = [response.datatypecollection.types];
+					}
+					// Need to retrieve dates now. The dates are the same for all variables so use the first in the list
+					self.add(_.map(dataCollection, function(c) {
+						return {
+							name : c.name,
+							description : c.description,
+							unitsstring : c.unitsstring
+						};
+					}));
+					deferred.resolve();
+				}
+				else {
+					deferred.reject(VARIABLE_WPS_PROCESS_ID + ' : Invalid wps response received');
+				}
+			}).fail(function(jqXHR, textStatus) {
+				deferred.reject(VARIABLE_WPS_PROCESS_ID + ' : Service request failed with ' + textStatus);
+			});
+
+			return deferred.promise();
+		}
     });
+
+	var DataSourceModel =  Backbone.Model.extend({
+		defaults : {
+			url : '',
+			variables : new DataSourceVariables(),
+			minDate : '',
+			maxDate : ''
+		},
+
+		getDateRange : function(options) {
+			// date range is the same for all variables, so just use the first one in the collection
+			var self = this;
+			var deferred = $.Deferred();
+			var wpsInputs = {
+				"catalog-url": [this.get('url')],
+				"allow-cached-response": [options.allowCached],
+				"grid": [this.get('variables').at(0).get('name')]
+			};
+			var wpsOutput = ["result_as_json"];
+
+			var hasExpectedDateProperties = function(obj) {
+				var expectedProperties = ['year','month','day'];
+				var hasExpectedNumericProperties = true;
+
+				if (_.isObject(obj)) {
+					var picked = _.pick(obj, expectedProperties);
+					if(_.keys(picked).length !== expectedProperties.length){
+						hasExpectedNumericProperties = false;
+					}
+					else {
+						var valuesAreNumeric = _.chain(picked).values().every(_.isNumber).value();
+						if(!valuesAreNumeric){
+							hasExpectedNumericProperties = false;
+						}
+					}
+				}
+				else{
+					hasExpectedNumericProperties = false;
+				}
+				return hasExpectedNumericProperties;
+			};
+
+			GDP.wpsClient.sendWpsExecuteRequest(
+				GDP.config.get('application').endpoints.utilityWps + '/WebProcessingService',
+				DATE_RANGE_WPS_PROCESS_ID,
+				wpsInputs,
+				wpsOutput,
+				false,
+				null,
+				true,
+				'json',
+				'application/json'
+			).done(function(response) {
+				var starttime, endtime;
+				if (_.has(response, 'availabletimes') && _.has(response.availabletimes, 'starttime') &&
+					_.has(response.availabletimes, 'endtime')  &&
+					hasExpectedDateProperties(response.availabletimes.starttime) &&
+					hasExpectedDateProperties(response.availabletimes.endtime)) {
+
+					starttime = response.availabletimes.starttime;
+					endtime = response.availabletimes.endtime;
+
+					self.set('minDate', starttime.month + '/'  + starttime.day + '/' + starttime.year);
+					self.set('maxDate', endtime.month + '/'  + endtime.day + '/' + endtime.year);
+
+					deferred.resolve();
+				}
+				else {
+					deferred.reject(DATE_RANGE_WPS_PROCESS_ID + ' : Invalid response received');
+				}
+			}).fail(function(jqXHR, textStatus) {
+				deferred.reject(DATE_RANGE_WPS_PROCESS_ID + ' : Service call failed: ' + textStatus);
+			});
+
+			return deferred.promise();
+		},
+
+		fetch : function(options) {
+			var self = this;
+			var deferred = $.Deferred();
+			var variables = new DataSourceVariables();
+			this.set('url', options.dataSourceUrl);
+
+			variables.fetch(options).done(function() {
+				self.set('variables', variables);
+				self.getDateRange({
+					allowCache : options.allowCache
+				}).done(function() {
+					deferred.resolve();
+				}).fail(function(msg) {
+					deferred.reject(msg);
+				});
+			}).fail(function(msg) {
+				deferred.reject(msg);
+			});
+			return deferred.promise();
+		}
+	});
 
     GDP.PROCESS_CLIENT.model.DataSourceVariable = DataSourceVariable;
     GDP.PROCESS_CLIENT.model.DataSourceVariables = DataSourceVariables;
+	GDP.PROCESS_CLIENT.model.DataSourceModel = DataSourceModel;
 }());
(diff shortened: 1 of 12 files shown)


mbucknell:
@cschroed-usgs , if you have time could you review this? I may merge before you get to it and if you can't that's fine.

src/main/webapp/js/process_client/models/DataSourceModels.js
```diff
@@ -1,5 +1,6 @@
 /*jslint browser: true*/
 /*global Backbone*/
```
cschroed-usgs: If you want, you can also put all of your jslint globals on one line:
http://www.jslint.com/help.html#global

src/main/webapp/js/process_client/models/DataSourceModels.js
```diff
@@ -7,18 +8,184 @@ var GDP = GDP || {};
...
-	defaults:{
-	    text: '',
-	    value: null,
-	    selected: false
-	}
-    });
+		defaults:{
+			name : '',
+			description : '',
+			unitsstring : ''
```
cschroed-usgs: I see a lot of other properties in this project that use camelCase. Should this property use it too?

src/main/webapp/js/process_client/models/DataSourceModels.js
```diff
@@ -7,18 +8,184 @@ var GDP = GDP || {};
...
 
     var DataSourceVariables = Backbone.Collection.extend({
-	model: DataSourceVariable
+		model: DataSourceVariable,
+
+		/*
+		 * @param {Object} options
+		 *     @prop {String} dataSourceUrl - The catalog url to be used to retrieve the data variables
+		 *     @prop {Boolean} allowCached
+		 *  @returns a promise that is resolved when the variables are succussfully retrieve, otherwise rejected
```
cschroed-usgs: When I return a promise I like to explicitly state what args the resolution is called with, even if it is called with no args. We could just assume that if the docs don't say what the args are, there are none, but that seems implicit rather than explicit, and python teaches us we should favor explicitness. What do you think? Is it reasonable to expect that explicit documentation?

src/main/webapp/js/process_client/models/DataSourceModels.js
```diff
@@ -7,18 +8,184 @@ var GDP = GDP || {};
...
+				var dataCollection;
+				if (response && _.has(response, 'datatypecollection') && _.has(response.datatypecollection, 'types')) {
+					if (_.has(response.datatypecollection.types, 'length')) {
+						dataCollection = response.datatypecollection.types;
+					}
+					else {
+						dataCollection = [response.datatypecollection.types];
+					}
+					// Need to retrieve dates now. The dates are the same for all variables so use the first in the list
+					self.add(_.map(dataCollection, function(c) {
```
cschroed-usgs: Just like there's an app for everything, there's an underscore method for everything :)

An underscore method called [_.pick](http://underscorejs.org/#pick) could shorten this common use case up a bit:
```javascript
self.add(
  _.map(dataCollection, function(c){
    return _.pick(c, 'name', 'description', 'unitsstring');
  })
);
```

_.pick also has an alter ego : [_.omit](http://underscorejs.org/#omit)

What do you think? Are you a fan of these underscore Object convenience methods? Got any favorites?

src/main/webapp/js/process_client/models/DataSourceModels.js
```diff
@@ -7,18 +8,184 @@ var GDP = GDP || {};
...
+				else {
+					deferred.reject(DATE_RANGE_WPS_PROCESS_ID + ' : Invalid response received');
+				}
+			}).fail(function(jqXHR, textStatus) {
+				deferred.reject(DATE_RANGE_WPS_PROCESS_ID + ' : Service call failed: ' + textStatus);
+			});
+
+			return deferred.promise();
+		},
+
```
cschroed-usgs: Could you please add a docblock here? Though it's a short function, it's handy to know what the promise rejection and resolution conditions are.

src/main/webapp/js/process_client/models/JobModel.js
```diff
@@ -89,12 +84,12 @@ var GDP = GDP || {};
 						}
 						else {
 							dataSetModel.clear();
-							deferred.reject();
```
cschroed-usgs: Should rejecting the promise with a message be documented in the doc block?

src/main/webapp/js/process_client/models/JobModel.js
```diff
@@ -216,9 +203,7 @@ var GDP = GDP || {};
 
 			/* The following property always need to be specified */
 			var result = {
-				DATASET_ID : _.map(this.getSelectedDataSourceVariables(), function(model) {
-					return model.get('value');
-				})
+				DATASET_ID : this.get('dataVariables')
```
cschroed-usgs: I get confused by this because the property name on the left is singular, but the parameter on the right is plural. Could they both be plural or could they both be singular?

src/main/webapp/js/process_client/views/DataDetailsView.js
```diff
@@ -24,405 +24,263 @@ var GDP = GDP || {};
...
+		 *     @prop {GDP.PROCESS_CLIENT.JobModel} model
+		 *     @prop {String} datasetId,
+		 */
+		'initialize': function(options) {
+			var self = this;
+			//Fetch data set model
+			var getDataSetModelPromise = this.model.updateDataSetModel(options.datasetId);
+			var hasVariables = this.model.get('dataSourceModel').get('variables').length > 0;
+
+			this.routePrefix = options.datasetId ? '#!catalog/gdp/dataset/' + options.datasetId : '#!advanced';
```
cschroed-usgs: Can backbone do named routes?

src/main/webapp/js/process_client/views/DataDetailsView.js
```diff
@@ -24,405 +24,263 @@ var GDP = GDP || {};
...
-				else {
-					//todo: anything better than 'alert'
-					var message = self.failedToParseVariableResponseMessage;
-					alert(message);
-					deferred.reject(message);
+
+			getDataSetModelPromise.done(function() {
+				var dataSourceModel = self.model.get('dataSourceModel');
+
+				self.listenTo(self.model, 'change:dataSourceUrl', self.changeUrl);
```
cschroed-usgs: Maybe there's an order dependence thing that I'm not grasping here, but I can't figure out why these event handlers are getting set during a promise resolution. Is it ok to put them in the view's 'events' property at the top?

cschroed-usgs: Leaving off here. Will continue review in the morning.

src/main/webapp/js/process_client/models/DataSourceModels.js
mbucknell: I actually like putting them on separate lines. It's easier to see what is global.

src/main/webapp/js/process_client/models/DataSourceModels.js
mbucknell: In this case, it matches what is received from the web service response.

src/main/webapp/js/process_client/models/DataSourceModels.js
mbucknell: I agree with explicit is better than implicit. However when nothing is returned, I think its fine to not explicitly say that nothing is returned.

src/main/webapp/js/process_client/models/DataSourceModels.js
mbucknell: I do like them. Missed this one. I'm becoming a bigger fain of _.chain

src/main/webapp/js/process_client/models/JobModel.js
mbucknell: Probably should be plural.

src/main/webapp/js/process_client/views/DataDetailsView.js
mbucknell: Backbone navigate expects a string ... it doesn't do 'named' routes.

src/main/webapp/js/process_client/views/DataDetailsView.js
mbucknell: The events property is for DOM events, these are model events. I put them here because I want to make sure the dataset information has been retrieved before setting event handlers on the Job Model.

src/main/webapp/js/process_client/models/DataSourceModels.js
cschroed-usgs: That's cool. Just making sure you knew about the other way too.

src/main/webapp/js/process_client/models/DataSourceModels.js
cschroed-usgs: Ahh, makes sense. Thanks for explaining.

src/main/webapp/js/process_client/models/DataSourceModels.js
cschroed-usgs: Oh yeah! I like _.chain too, but had forgotten about it. Thanks for the re-mind :)

src/main/webapp/js/process_client/models/JobModel.js
cschroed-usgs: Sweet!

src/main/webapp/js/process_client/views/DataDetailsView.js
cschroed-usgs: Shucks. Named routes are super. I googled around for some plugins for Backbone named routes. Nothing promising came up -- just a bunch of alpha stuff. Since there's no reliable relevant tooling yet, this is cool as-is.

src/main/webapp/js/process_client/views/DataDetailsView.js
cschroed-usgs: Gotcha. Makes sense. Thanks for explaining!

src/main/webapp/js/process_client/views/DataDetailsView.js
```diff
@@ -24,405 +24,263 @@ var GDP = GDP || {};
...
+		 * @param {Jquery event} ev
+		 * @returns {undefined}
+		 */
+		goToHubPage : function(ev) {
+			ev.preventDefault();
+			this.router.navigate(this.routePrefix, {trigger : true});
+		},
+
+		setInputsDisabled : function(disabled) {
+			this.$(urlTextPicker.selector).prop('disabled', disabled);
```
cschroed-usgs: jQuery is accessible via the file-wrapping IIFE arg '$' from line 4. Do you know why jQuery is being accessed via a '$' property on 'this'?

cschroed-usgs: Never mind! Found the answer in the docs:
http://backbonejs.org/#View-dollar
That is good to know!

src/main/webapp/js/process_client/views/DataDetailsView.js
```diff
@@ -24,405 +24,263 @@ var GDP = GDP || {};
...
+				dataSourceModel.fetch({
+					dataSourceUrl : dataSourceUrl,
+					allowCached : this.$('#use-cached-checkbox').is(':checked')
+				}).fail(function(msg) {
+					self.model.set('dataVariable', []);
+					self.model.set('startDate', '');
+					self.model.set('endDate', '');
+					//TODO: better error handling
+					GDP.logger.debug(msg);
+					window.alert(msg);
```
cschroed-usgs: Is better error handling in the scope of this ticket?

cschroed-usgs: I noticed below that the HubView uses the following to do user-facing logging. Is that easy to do from here?
```javascript
self.childViews.alertView.show('alert-info', 'Process status: started');
```

mbucknell: Answer to first question is no. Answer to the second question is that these messages are directly related to the processing. We show status as we are waiting as well as warnings, errors, and a succeeded message.

Error handling that is visible to the user should be in the view, so I'm going to take out this window.alert here. I think we will need to open a technical debt ticket to improve error handling.

src/test/js/specs/DataSourceModels_spec.js
```diff
@@ -0,0 +1,254 @@
...
+							unitsstring : 'Unit3'
+						}
+					]
+				}
+			};
+			testCollection.fetch({
+				dataSourceUrl : 'http://fakeDataSource',
+				allowCached : false
+			}).done(successSpy).fail(failedSpy);
+
```
cschroed-usgs: Why do we expect that the success and failure spies will not be called? Shouldn't one of them be called?

src/main/webapp/js/process_client/models/JobModel.js
mbucknell: Done

src/main/webapp/js/process_client/models/JobModel.js
mbucknell: Actually this has to be DATASET_ID as it is feed to the wps process request. And if you notice all of them are arrays anyways.

src/test/js/specs/DataSourceModels_spec.js
mbucknell: Because the promise that is returned by fetch has not yet been resolved. That is done in line 94.

src/main/webapp/js/process_client/views/DataDetailsView.js
cschroed-usgs: Sounds good to me!

src/main/webapp/js/process_client/models/JobModel.js
cschroed-usgs: :+1:

src/main/webapp/js/process_client/models/JobModel.js
cschroed-usgs: Ok. That makes sense. It might be helpful to leave a comment there explaining the plurality mismatch.

mbucknell:
@cschroed-usgs I'm going to merge because I've got updates for another ticket.

src/test/js/specs/DataSourceModels_spec.js
cschroed-usgs: Ahhh, roger that.

cschroed-usgs:
:+1:

PR #115, opened 2015-09-22
Fixed bug when setting the redirect_url when ther are no incomingParams.

mbucknell:


mbucknell:
@jiwalker-usgs and @dblodgett-usgs , this should fix the emailer issue. Should I cut a release and push to prod?

jiwalker-usgs:
@mbucknell Yeah, I'd vote to :shipit:

PR #116, opened 2015-10-06
We are no longer creating the wpsClient and passing it in to views. I…

mbucknell:
… missed the instance of this in SpatialView. Now using GDP.wpsClient.


mbucknell:
@dblodgett-usgs , this will fix the issue with the draw polygon feature on the spatial page.

PR #117, opened 2015-10-06
Gdp 1005

cschroed-usgs:


src/main/java/gov/usgs/cida/gdp_web/service/GDPSearchCrawlerServicer.java
```diff
@@ -0,0 +1,62 @@
+package gov.usgs.cida.gdp_web.service;
```
cschroed-usgs: Except for the package name and the class name, this file is identical to the [servicer in NWC](https://github.com/USGS-CIDA/nwc/blob/13de346338d665f30eba0ca236d3861b37dc577f/nwc/src/main/java/gov/usgs/cida/nwc/servlet/NWCSearchCrawlerServicer.java).

Part of me wants to incorporate this as a SHA1SearchCrawlerServicer into the ajax-search-crawler-tools repo to avoid duplication, but I'm not sure that it is a good idea yet because two instances of duplication do not necessarily constitute a trend. What do you think? Move it in to the shared module, or wait and see?

mbucknell: I'm all for duduping, however would putting it in ajax-search-crawler-tools be the right place?

jiwalker-usgs: I would suggest implementing SHA1SearchCrawlerServicer as reference implementation in ajax-search-crawler-tools and then have the GDP and NWC subclass this with any differences that come up.

jiwalker-usgs: Also, it is probably just me, but the underscores in the package name trigger my spidey sense a bit.

mbucknell:
:+1:

src/main/java/gov/usgs/cida/gdp_web/service/GDPSearchCrawlerServicer.java
cschroed-usgs: Thanks for the feedback guys. I will move this SearchCrawlerServicer into the ajax-search-crawler-tools module.

cschroed-usgs: @jiwalker-usgs , I looked around for specs and best practices on Java package naming. I couldn't find anything definitive about underscores in the package name after the organization's domain name. If there are non-alphanumeric chars in an org's website name, [this oracle doc](https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html) says it's cool to do use underscores to represent the bad chars. Is there something I'm missing?

jiwalker-usgs: No, I don't think it is technically an issue, my eyes just don't like it for some reason.

cschroed-usgs: haha. Ok. I'm ever-so-slightly biased towards non-alphanumeric word boundaries because it makes it easier to read names for people who don't use :gb: as a first language. I'm happy to subordinate that to a community standard or best practice though if we ever find something anti-underscore out there.

PR #132, opened 2016-06-04
Re - defaulting variable cache use and removing the auto-select if only one dataset. 

dblodgett-usgs:
This is a reasonable solution to the problem with controlling the use of the cache. We may want to revisit this workflow so it's more intuitive and easy in the future, but this works well for now.


dblodgett-usgs:
@mbucknell - Can you make sure I didn't mess anything up here? I built and tested this on my local host but don't know if there are tests or any other code that I don't know about and should have updated.

dblodgett-usgs:
Going to merge now for testing on dev/qa

mbucknell:
👍

PR #146, opened 2017-03-02
GDP Welcome Page Update

mwernimont:
@mbucknell @jiwalker-usgs here's the code for the updated Welcome page.

I'm expecting a few comments on the welcome.js since I had some scoping issues that I did my best to solve, but let me know what could be improved.


src/main/webapp/js/landing/welcome.js
```diff
@@ -0,0 +1,88 @@
+ //Ran into Scope issues and this is the way I could find the elements I wanted events on
```
mbucknell: I think the problem you where having was setting up the event handlers before the page is loaded. If you look in landing/init.js on line 10 is the jquery command that waits until the page is loaded before executing the code. You could either do the same thing in this file.  Is this a new page or is it replacing the current Welcome page? If it is new, you might want to consider putting the file in a new directory rather than in the landing directory (or perhaps rename landing to a better name).

src/main/webapp/js/landing/welcome.js
```diff
@@ -0,0 +1,88 @@
...
+        // can just use the original source element for the selection and copy
+        target = elem;
+        origSelectionStart = elem.selectionStart;
+        origSelectionEnd = elem.selectionEnd;
+    } else {
+        // must use a temporary form element for the selection and copy
+        target = document.getElementById(targetId);
+        if (!target) {
+            var target = document.createElement("textarea");
+            target.style.position = "absolute";
```
mbucknell: I wonder if you can just set target.style.display = "none" to hide the element.

src/main/webapp/js/landing/welcome.js
mbucknell: On further thought, these event handlers need to go into the appropriate Backbone.View which is in js/util/WelcomeView. This will also change the welcome banner on the client page, so we may want to create a new view. Maybe we can talk about this.

mwernimont:
@mbucknell look better to you?

mbucknell:
Code looks better. You should consider trying to clean up the less files and remove styles that are no longer used.

mwernimont:
Good suggestion, less code the better

mwernimont:
@mbucknell do you have time today to review my last commit and see if this PR is mergable? Would like for this to count for my sprint that ends tomorrow if I could.

PR #147, opened 2017-03-08
fixed tests

mwernimont:
Removed most of the tests that do not apply anymore and have one checking my toggleWelcomeArea function on the area and the image having a class.

Look alright to at least get dev building again? @cschroed-usgs @mbucknell ?


mwernimont:
screen shot 2017-03-08 at 1 01 19 pm

Got a positive result from the jasmine testing area Mary showed me.

mbucknell:
I think this is good to go.

PR #148, opened 2017-03-27
landing page fixes

mwernimont:
Images appear on both pages the welcome template is pulled to.

Welcome view is hidden initially when not on the landing page.

@mbucknell I was having difficulties getting what I wanted to happen on the welcome view, so I wouldn't be surprised if you had a better solution than what I came up with. So if you have any suggestions please let me know.

### DIFF
--- a/src/main/webapp/js/util/WelcomeView.js
+++ b/src/main/webapp/js/util/WelcomeView.js
@@ -8,7 +8,7 @@ GDP.util = GDP.util || {};
 	"use strict";
         
 	GDP.util.WelcomeView = GDP.util.BaseView.extend({
-
+            
 		events : {
                         'click #copyCitation' :  'copyToClipboard',
                         'click .welcomeToggleButton' : 'toggleWelcomeArea', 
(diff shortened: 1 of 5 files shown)


src/main/webapp/js/util/WelcomeView.js
```diff
@@ -80,14 +69,15 @@ GDP.util = GDP.util || {};
...
+                
 		toggleWelcomeArea : function() {
+                    var button = $('.welcomeToggleButton');
                     //Hides Welcome Content
                     $('.welcome-content').slideToggle('slow');
                     //Removes margin from top to rest neatly against header
                     $('.toggleButtonArea').toggleClass('noMargin');
                     //Triggers animation for toggle button
-                    $('.welcomeToggleButton').toggleClass('animationTrigger');
+                    $(button).toggleClass('animationTrigger');
```
mbucknell: Not sure why you are assigning $('.welcomeToggleButton') to a variable and then only using it once. So I think this can be simplified. Also for feature reference the line could be written as button.toggleClass('animationTrigger');

src/main/webapp/templates/welcome.html
```diff
@@ -19,29 +19,28 @@
                     <div class="gdpOption">
                         <h4>PYGDP</h4>
                         <div class="optionImage">
-                            <img src="images/python.svg" alt="Python Language Logo"/>
+                            <img id="python" src="images/python.svg" alt="Python Language Logo"/>
```
mbucknell: Rather the changing the src attribute in the view, you could pass a context variable for the path to the images folder to the template. So this line would become something like Python Language Logo. You would need to change WelcomeView's render to pass in a context containing {pathToImages: GDP.BASE_URL}

This is cleaner in that you should avoid making changes to a child view from a parent view.

mwernimont:
@mbucknell Got the context to work for the render, a lot cleaner like you said.

PR #151, opened 2017-04-04
text color and margin

mwernimont:
@jiwalker-usgs Text is black and we got the spacing back for those divs to stay away from the header.


jiwalker-usgs:
👍

PR #157, opened 2018-07-10
add red river dataset to algorithms.json

wdwatkins:


wdwatkins:
It is clean.  I don't have merge privileges, can you merge @dblodgett-usgs?

PR #162, opened 2020-01-04
release Hawaii dataset

wdwatkins:
New branch that doesn't include datasets on master that aren't ready yet


src/main/webapp/WEB-INF/jsp/algorithms.json
```diff
@@ -188,21 +188,13 @@
...
-	],
-	"5977b95de4b0ec1a4888e493": [
-	    "gov.usgs.cida.gdp.wps.algorithm.FeatureWeightedGridStatisticsAlgorithm",
-	    "gov.usgs.cida.gdp.wps.algorithm.FeatureCoverageOPeNDAPIntersectionAlgorithm"
+	"5977b95de4b0ec1a4888e493": [	
+	    "gov.usgs.cida.gdp.wps.algorithm.FeatureWeightedGridStatisticsAlgorithm",	
+	    "gov.usgs.cida.gdp.wps.algorithm.FeatureCoverageOPeNDAPIntersectionAlgorithm"	
 	],
-	"5cba5abce4b0c3b0066185cc": [
+	"5a2710b6e4b03852bafb5b21": [
```
wdwatkins: Net change here is removing two datasets, additions are just because the order has changed

